PR #1, opened 2015-04-26
Added configparser as install requirement

idigon:
After installing commad "todo" requested configparser so I added it as a requiremtn

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,7 +23,8 @@ setup(
         'icalendar',
         'urwid',
         'pyxdg',
-        'atomicwrites'
+        'atomicwrites',
+	'configparser'
     ],
     # TODO: classifiers
 )


hobarrera:
[configparser](https://docs.python.org/3/library/configparser.html) is included as part of python since python 3.2. Older versions of python are not supported and todoman has other issues with those versions.

hobarrera:
(Keep in mind that python2 and python3 can happily coexist if you have python2 as the default. Just use the latest source from git).

idigon:
I have used Python2 with the latest source from git and needed configparser, that is why I added it. Anyway, you specifically state it works with Python3, which indeed includes configparser, I will try it with that version

hobarrera:
> I have used Python2 with the latest source from git and needed configparser, that is why I added it. Anyway, you specifically state it works with Python3, which indeed includes configparser, I will try it with that version

Yup. Looks like I hadn't pushed the latest source to github (where this is explicitly stated), sorry about that. There are a few other issues with python2 which you'd probably hit over time too.

idigon:
I still use some projects which require Python2 so I guess I will stick with it for a while although, as you say, both can cohexist

hobarrera:
> I still use some projects which require Python2 so I guess I will stick with it for a while although, as you say, both can cohexist

Yup. You don't need to change your default python or affect you python2 installation in any way. :)
Lemme know if you still encounter any issues though.

Cheers

PR #2, opened 2015-04-26
Fix database init

idigon:
Database __init__ method did not call _read to populate todos list

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,7 +23,8 @@ setup(
         'icalendar',
         'urwid',
         'pyxdg',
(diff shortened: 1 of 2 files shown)


hobarrera:
This is done by the setter for `path`. Python 2 wasn't calling the setter (I believe the class needs to extend `object` for that to happen). However, this is no longer applicable, as per #1.

idigon:
Yeah, again, Python2 vs Python3. Sorry, man!

hobarrera:
No problem! ^_^

PR #3, opened 2016-04-13
Implement custom sorting cli option

hobarrera:
Implement `todo list --sort FIELD1,FIELD2`, to allow custom sorting.

Though the code seems complex, there's no noticeable difference between this sorting mechanism and the original one, though I'd love some feedback on how to de-uglify this.

Code can also be used to make the default sort order configurable.

### DIFF
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -169,15 +169,30 @@ def test_sorting_output(tmpdir, runner, create):
         'SUMMARY:bbb\n'
         'DUE;VALUE=DATE-TIME;TZID=ART:20160101T000000\n'
     )
-    result = runner.invoke(cli, ['list', '--sort', '-summary'])
-    assert not result.exception
-    assert 'aaa' in result.output.splitlines()[0]
-    assert 'bbb' in result.output.splitlines()[1]
 
-    result = runner.invoke(cli, ['list', '--sort', 'due'])
-    assert not result.exception
-    assert 'aaa' in result.output.splitlines()[0]
-    assert 'bbb' in result.output.splitlines()[1]
+    examples = [
+        ('-summary', ['aaa', 'bbb']),
+        ('due', ['aaa', 'bbb'])
+    ]
+
+    # Normal sorting, reversed by default
+    all_examples = [(['--sort', key], order) for key, order in examples]
+
+    # Testing --reverse, same exact output
+    all_examples.extend((['--reverse', '--sort', key], order)
+                        for key, order in examples)
+
+    # Testing --no-reverse
+    all_examples.extend((['--no-reverse', '--sort', key], reversed(order))
+                        for key, order in examples)
+
+    for args, order in all_examples:
+        result = runner.invoke(cli, ['list'] + args)
+        assert not result.exception
+        lines = result.output.splitlines()
+        for i, task in enumerate(order):
+            assert task in lines[i]
+
 
 # TODO: test aware/naive datetime sorting
 # TODO: test --grep
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -113,7 +113,6 @@ def test_delete(tmpdir, runner, create):
     assert len(result.output.splitlines()) == 0
 
 
-
 def test_dtstamp(tmpdir, runner, create):
     """
     Test that we add the DTSTAMP to new entries as per RFC5545.
@@ -127,7 +126,7 @@ def test_dtstamp(tmpdir, runner, create):
     assert todo.dtstamp.tzinfo is pytz.utc
 
 
-def test_sorting(tmpdir, runner):
+def test_sorting_fields(tmpdir, runner):
     for i in range(1, 10):
         days = datetime.timedelta(days=i)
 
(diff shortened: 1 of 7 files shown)


todoman/main.py
```diff
@@ -30,25 +30,45 @@ def dump_idfile(ids):
...
+                # Remove that '-'
+                field = field[1:]
+            if field == 'due':
+                # TODO: Some better method to deal with these special cases?
+                value = todo.due.timestamp() if todo.due else float('inf')
+            elif field == 'created_at':
+                value = todo.created_at.timestamp() if todo.created_at else 0
+            else:
+                value = getattr(todo, field, "")
+            value = -value if neg and value else value
```
untitaker: Haven't tested, but should crash if given `-summary`.

hobarrera: Yes, it does.

I can't really think of any other way to keep this generic *and* working for al these edge cases.

untitaker: Fixed!

hobarrera:
Thanks for adding to this, @untitaker. I'll try to add some basic tests that make sure that at least a few pattern works to catch future regressions and merge this into master.

untitaker:
Which tests are missing? Are you talking about ones that actually assert correct output?

hobarrera:
> Which tests are missing? Are you talking about ones that actually assert correct output?

Yes, at least a basic one to catch regression. I've just added one and think that's enough.

hobarrera:
As-is, sorting seems to be the opposite of what one would naturally expect.

Sorting by summary lists:
```
ccc
bbb
aaa
```

Sorting by due date:
```
2015-11-11
2015-06-16
2015-05-31
```

IMHO, we should invert this. Opinions?

untitaker:
Didn't we invert this on purpose such that you don't have to scroll to see the topmost entries?

On 17 April 2016 05:26:54 CEST, Hugo Osvaldo Barrera <notifications@github.com> wrote:
>As-is, sorting seems to be the opposite of what one would naturally
>expect.
>
>Sorting by summary lists:
>```
>ccc
>bbb
>aaa
>```
>
>Sorting by due date:
>```
>2015-11-11
>2015-06-16
>2015-05-31
>```
>
>IMHO, we should invert this. Opinions?
>
>---
>https://github.com/pimutils/todoman/pull/3#issuecomment-210946504

-- 

untitaker:
Perhaps we should add a reverse flag in addition, such that you can pipe to a pager and reverse without having to modify your sort key.

On 17 April 2016 05:26:54 CEST, Hugo Osvaldo Barrera <notifications@github.com> wrote:
>As-is, sorting seems to be the opposite of what one would naturally
>expect.
>
>Sorting by summary lists:
>```
>ccc
>bbb
>aaa
>```
>
>Sorting by due date:
>```
>2015-11-11
>2015-06-16
>2015-05-31
>```
>
>IMHO, we should invert this. Opinions?
>
>---
>https://github.com/pimutils/todoman/pull/3#issuecomment-210946504

-- 

hobarrera:
I like this idea. It also makes sense for `todo list` in general.

untitaker:
So, `--reverse/--no-reverse` with default=True?

On 17 April 2016 10:32:18 CEST, Hugo Osvaldo Barrera <notifications@github.com> wrote:
>I like this idea. It also makes sense for `todo list` in general.
>
>---
>https://github.com/pimutils/todoman/pull/3#issuecomment-210976225

-- 

hobarrera:
> So, `--reverse/--no-reverse` with default=True?

Yes, yes. We'll need to reverse the default order though, so that `reverse=True` does not alter the output of `todo list` (eg: so that it continues printing everything in the same order).

untitaker:
That's what i meant, so you pass --no-reverse if you want to pipe to a pager.

On 17 April 2016 10:38:16 CEST, Hugo Osvaldo Barrera <notifications@github.com> wrote:
>> So, `--reverse/--no-reverse` with default=True?
>
>Yes, yes. We'll need to reverse the default order though, so that
>`reverse=True` does not alter the output of `todo list` (eg: so that it
>continues printing everything in the same order).
>
>---
>https://github.com/pimutils/todoman/pull/3#issuecomment-210977199

-- 

untitaker:
I added the --reverse/--no-reverse option with tests.

tests/test_basic.py
```diff
@@ -155,15 +155,30 @@ def test_sorting_output(tmpdir, runner, create):
...
-    result = runner.invoke(cli, ['list', '--sort', '-summary'])
-    assert not result.exception
-    assert 'aaa' in result.output.splitlines()[0]
-    assert 'bbb' in result.output.splitlines()[1]
 
-    result = runner.invoke(cli, ['list', '--sort', 'due'])
-    assert not result.exception
-    assert 'aaa' in result.output.splitlines()[0]
-    assert 'bbb' in result.output.splitlines()[1]
+    examples = [
```
untitaker: This is probably a bit overengineered now, but meh

untitaker:
I have no idea why the tests fail, they work for me outside of tox?

hobarrera:
> I added the --reverse/--no-reverse option with tests.

Thanks. :)

> I have no idea why the tests fail, they work for me outside of tox?

I've no idea - but they pass for me outside tox as well, and am trying to figure out what's going on.

untitaker:
I have the solution, I actually have a massive bug in the tests that creates corrupt todos that are skipped. Combine with broken logging for invalid todos and you have this behavior :)

hobarrera:
> I have the solution, I actually have a massive bug in the tests that creates corrupt todos that are skipped. Combine with broken logging for invalid todos and you have this behavior :)

Since you're using the `Todo` class, I'd expect them to be valid. Maybe also wrap creation using a `Database` class?

hobarrera:
Seems to work fine. I'll do the rebasing/merge once CI finishes, if  you think it's okay.

untitaker:
@hobarrera I'm using `Todo` from icalendar, which seems to behave very funky when doing `todo['value'] = <datetime>`. Fixed to use the todoman one. I'll also add a commit to not directly use icalendar at all.

PR #6, opened 2016-04-13
Add basic editing shortcuts to text widgets

hobarrera:
Some still-pending shortcuts:

 * <kbd>ctrl</kbd>+<kbd>s</kbd>: save (and exit?)
 * <kbd>tab</kbd>: next element (next input block or button)
 * <kbd>alt</kbd>+<kbd>f</kbd>: forward word
 * <kbd>alt</kbd>+<kbd>b</kbd>: backward word

I'd also like to possibly move these widgets to a separate project ("urwid-extras"?) so I can stop copy-pasting stuff from khal.

@geier: Thoughts?

### DIFF
--- a/todoman/widgets.py
+++ b/todoman/widgets.py
@@ -1,0 +1,103 @@
+# Copyright (c) 2016 Hugo Osvaldo Barrera
+# Copyright (c) 2013-2016 Christian Geier et al.
+#
+# Permission is hereby granted, free of charge, to any person obtaining
+# a copy of this software and associated documentation files (the
+# "Software"), to deal in the Software without restriction, including
+# without limitation the rights to use, copy, modify, merge, publish,
+# distribute, sublicense, and/or sell copies of the Software, and to
+# permit persons to whom the Software is furnished to do so, subject to
+# the following conditions:
+#
+# The above copyright notice and this permission notice shall be
+# included in all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+# EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+# MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+# NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+# LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+# OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+# WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+import re
+
+import urwid
+
+
+class ExtendedEdit(urwid.Edit):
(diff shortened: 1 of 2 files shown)


untitaker:
I'm unsure whether there isn't already a package or at least democode that already does this. Some readline-clone in Urwid?

hobarrera:
> I'm unsure whether there isn't already a package or at least democode that already does this. Some readline-clone in Urwid?

I haven't found one.  Nor an "urwid-extras" package, which really surprises me (though [it seems](https://github.com/urwid/urwid/issues/53) they're open to having one?).

geier:
I'm not sure if it makes sense yet to move that code to its own repo (I'm hesistant because of the added burden we put on package maintainers). Perhaps we should ask at urwid if there is any interest in a new urwid-contrib repository.

For a readline clone see https://github.com/urwid/urwid/pull/34 and there must be one in https://github.com/pazz/alot/

benjaminfrank:
this looks really nice :+1: 
maybe the keys could be configurable via the todoman config file and default to what you have now? as emacs user i like those bindings, but other might prefer something else

geier:
@hobarrera for tabbing have a look at NPile, NColumns and NlistBox in khal/ui/widgets.py (but some people have called it over-engineered, in truth I was just lazy).

untitaker:
Looks good though I don't see much value in separating functions like `goto_end_of_line` from ExtendedEditor.

hobarrera:
I've integrated their logic into the `ExtededEdit` methods.  
Some are even cleaner, while others might have room from improvements.

todoman/widgets.py
```diff
@@ -0,0 +1,103 @@
...
+            f_text = t[0:t.rfind('\n') + 1]
+        self.set_edit_text(f_text + text[self.edit_pos:])
+        self.set_edit_pos(len(f_text))
+
+    def _delete_till_end_of_line(self):
+        """delete till end of line before cursor"""
+        text = self.get_edit_text()
+        line = text[self.edit_pos:]
+
+        if line.find('\n') == -1:
```
untitaker: That's just `"\n" in line`, though you could store the result of `line.find` in a variable to reuse it later

todoman/widgets.py
```diff
@@ -0,0 +1,103 @@
...
+        self.set_edit_pos(len(f_text))
+
+    def _delete_till_beginning_of_line(self):
+        """delete till start of line before cursor"""
+        text = self.get_edit_text()
+        t = text[:self.edit_pos]
+        if t.rfind('\n') == -1:
+            f_text = ''
+        else:
+            f_text = t[0:t.rfind('\n') + 1]
```
untitaker: The `0` should not be necessary.

todoman/widgets.py
```diff
@@ -0,0 +1,103 @@
...
+        else:
+            f_text = t[:len(t) - len(words[-1])]
+
+        self.set_edit_text(f_text + text[self.edit_pos:])
+        self.set_edit_pos(len(f_text))
+
+    def _delete_till_beginning_of_line(self):
+        """delete till start of line before cursor"""
+        text = self.get_edit_text()
+        t = text[:self.edit_pos]
```
untitaker: I don't think this is necessary, you should be able to use the `start` and `end` parameters of `find` to avoid needless string slicing. (this is more of a performance improvement, not really for prettyness)

hobarrera:
Thanks for those three observations. All have been fixed now.

untitaker:
LGTM!

PR #7, opened 2016-04-13
Hypothesis-based test for --sort

untitaker:


untitaker:
This test should fail (and does locally). Maybe they're running too slowly? Added CI profile.

untitaker:
Also added bugfix. @hobarrera please review

PR #8, opened 2016-04-13
Refactor --sort logic, nicer error messages

untitaker:


todoman/main.py
```diff
@@ -39,6 +43,10 @@ def get_task_sort_function(fields):
             '-created_at',
         ]
 
+    for field in fields:
+        if not hasattr(model.Todo, field):
```
hobarrera: This would seem to fail for "negative fields", eg: `-priority`.

untitaker: Yes, just noticed.

hobarrera:
LGTM. Feel free to merge when tests complete.

untitaker:
Ah, this is buggy as well if the property is None.

untitaker:
Turns out the logic isn't that fixable, we still have the hardcoded specialcases.

PR #30, opened 2016-04-30
Implement color option

### LINKED ISSUE #29: Filter out colors if piped to other command
Currently, `todo list | cat` outputs colors. I think it shouldn't, and also should get a `--color` option (see khal).

Opinions?

untitaker:
Fix #29


hobarrera:
Thanks! :+1:

hobarrera:
(I'm okay with no tests since click should be testing this; we're just passing a flag).

PR #32, opened 2016-05-07
Add shortcut to edit textfield in editor

untitaker:


untitaker:
Try hitting <key>Ctrl + E</key> in a textfield.

If you agree I'll add docs.

todoman/ui.py
```diff
@@ -74,9 +96,16 @@ def edit(self):
...
-        loop = urwid.MainLoop(self._ui, unhandled_input=self._keypress,
-                              handle_mouse=False)
-        loop.run()
+        self._loop = urwid.MainLoop(self._ui, unhandled_input=self._keypress,
+                                    handle_mouse=False)
+        try:
+            self._loop.run()
+        except Exception:
+            try:  # Try to leave terminal in usable state
+                self._loop.stop()
```
untitaker: This was taken from khal, where I also added it. The urwid docs don't indicate it's necessary, however, I found https://github.com/urwid/urwid/issues/77

hobarrera:
> Try hitting <kbd>Ctrl + E</kbd> in a textfield.

Terminals commonly use <kbd>Ctrl + A</kbd> and <kbd>Ctrl + E</kbd> to move to start-of-line and end-of-line respectively.
I like the idea, but I'd also like to actually have those keybinding do the same here in future (like <kbd>Ctrl + W</kbd>, etc).

Again, I really like the idea, so I think just changing the key combination would be fine.

untitaker:
To what?

untitaker:
I changed it to `ctrl-i`. This is still a used key within Emacs, but not in Bash.

untitaker:
Just saw #6, probably better to put this on ice until that one is finished.

untitaker:
@hobarrera Any update on this and #6? Thanks!

hobarrera:
Does #6 look okay to you? If it does, I'll integrate this branch into it, and then the whole thing into master.

PR #40, opened 2016-10-03
Add `default_list` in config for new todos

nim65s:
Create todos fastly and easily.

Maybe, if  there is only one list, it should automaticaly considered as the default one.

### DIFF
--- a/todoman.conf.sample
+++ b/todoman.conf.sample
@@ -2,3 +2,4 @@
 # A glob expression which matches all directories relevant.
 path = ~/.local/share/calendars/*
 date_format = %Y-%m-%d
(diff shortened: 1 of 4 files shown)


todoman.conf.sample
```diff
@@ -2,3 +2,4 @@
 # A glob expression which matches all directories relevant.
 path = ~/.local/share/calendars/*
 date_format = %Y-%m-%d
+default_list = Personnal
```
untitaker: Single `n`

hobarrera:
LGTM, but please jnclude at least one test for this.

nim65s:
Is this enough ?

PR #43, opened 2016-10-04
Show error messages in UI

### LINKED ISSUE #42: UI: swallows invalid due dates
If I edit a task, enter something completely ridiculous into the due-field and hit save, todoman will silently discard all changes I've made.

untitaker:
Fix #42

Also:

- Introduce a color palette
- Change the errorhandling around MainLoop to actually show tracebacks
  instead of just exiting

### DIFF
--- a/todoman/ui.py
+++ b/todoman/ui.py
@@ -1,5 +1,3 @@
-import sys
-
 from datetime import datetime
 from time import mktime
 
@@ -98,13 +96,12 @@ class TodoEditor:
         )
         try:
             self._loop.run()
-        except Exception as e:
-            tb = sys.exc_info()[-1]
+        except Exception:
             try:  # Try to leave terminal in usable state
                 self._loop.stop()
             except Exception:
                 pass
-            raise e.with_traceback(tb)
+            raise
         self._loop = None
         return self.saved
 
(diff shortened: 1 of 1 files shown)


untitaker:
This one too @hobarrera

hobarrera:
Oh, I wasn't aware that I needed to confirm my comments as a review before they appeared now.

todoman/ui.py
```diff
@@ -70,31 +78,46 @@ def __init__(self, todo, databases, formatter):
...
         try:
             self._loop.run()
-        except Exception:
+        except Exception as e:
+            tb = sys.exc_info()[-1]
             try:  # Try to leave terminal in usable state
                 self._loop.stop()
             except Exception:
                 pass
+            raise e.with_traceback(tb)
```
untitaker: If stopping the loop raises an exception it would be suppressed in the inner try-except and reraised here. This is not the behavior we want, we want to reraise the original exception before we tried to stop the loop.

It appears that this has changed for Python 3 and `raise` is now somehow aware of scoped trys.

untitaker: Fixed.

untitaker:
Your review is also sent as a single email after you submit it.

hobarrera:
LGTM.

PR #50, opened 2016-10-25
Adding a copy command

nim65s:
to copy an item into another list.

I usually like to copy items from public read-only calendars to the one I share with my relatives, so I would really appreciate this little functionality.

If you agree with this PR, I'll probably make another similar one for moving items from an agenda to another (ie. copy + delete). This can be useful when someone mistakenly creates a new item in a wrong list.

### DIFF
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -113,6 +113,24 @@ def test_delete(tmpdir, runner, create):
     assert len(result.output.splitlines()) == 0
 
 
+def test_copy(tmpdir, runner, create):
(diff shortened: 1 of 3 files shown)


hobarrera:
Looks great, thanks!

hobarrera:
> If you agree with this PR, I'll probably make another similar one for moving items from an agenda to another (ie. copy + delete). This can be useful when someone mistakenly creates a new item in a wrong list.

I've done that a few times, so that's be helpful, thanks.

PR #52, opened 2016-10-26
Share parameters between new and edit

untitaker:
This should make it easier to add new options for VTODO props that are
then automatically shared between new and edit.

Also `todo edit` is no longer interactive by default.


hobarrera:
LGTM (I think I was this on mobile a week ago, and didn't reply, apparently).

PR #53, opened 2016-10-30
Remove ansi dependency

untitaker:
Most colorcodes are provided by Click, use Khal's colorcode conversion.
This yields different colors than the old algorithm, but provides more
accurate results for modern terminals.

### DIFF
--- a/todoman/model.py
+++ b/todoman/model.py
@@ -5,7 +5,6 @@ from os.path import normpath, split
 from uuid import uuid4
 
 import icalendar
-from ansi.colour.rgb import rgb8
 from atomicwrites import AtomicWriter
 from dateutil.tz import tzlocal
 
@@ -305,7 +304,7 @@ class Database:
     def color_ansi(self):
         rv = self.color_rgb
         if rv:
-            return rgb8(*rv)
+            return '\33[38;2;{!s};{!s};{!s}m'.format(*rv)
 
     def __str__(self):
         return self.name
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,5 +3,4 @@ icalendar
 urwid
 pyxdg
 atomicwrites
-ansi>=0.1.3
 parsedatetime
--- a/todoman/ui.py
+++ b/todoman/ui.py
@@ -1,9 +1,7 @@
 from datetime import datetime
 from time import mktime
+import click
 
-import ansi.colour.fg
-import ansi.colour.fx
-import ansi.sequence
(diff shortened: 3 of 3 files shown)


todoman/model.py
```diff
@@ -305,7 +304,7 @@ def color_rgb(self):
     def color_ansi(self):
         rv = self.color_rgb
         if rv:
-            return rgb8(*rv)
+            return '\33[38;2;{!s};{!s};{!s}m'.format(*rv)
```
hobarrera: Isn't is possible to use `click.style` here as well?  
This ends up getting printed by `TodoFormatter.compact()`.

hobarrera:
This also unbroke some colour issues I'd been having and didn't manage to track down. Thanks!

todoman/model.py
untitaker: `click.style` can't handle RGB colors, though it could.

untitaker: >This ends up getting printed by TodoFormatter.compact().

Not sure why that changes anything? What do you mean?

untitaker: Also we're supposed to return a color-code sequence here, but `click.style` wraps a given string with colorcode and reset code.

untitaker:
BTW I feel like we've investigated once to directly use RGB-colorcodes in the terminal and instead settled for `ansi`'s algorithm to round down to the next terminal color because the RGB-variant didn't work. I wonder why it does now.

untitaker:
Here's the original commit: https://github.com/pimutils/todoman/commit/c9e9b36 I think this was on GitLab though.

hobarrera:
Honestly, I think at the time 24bit-colour support wasn't were it was today. `urxvt` definitely did not support those colours back then, so maybe that's why we stuck with `ansi`'s algorithm. A lot has advanced in the last year or so regarding truecolour support in terminals.

todoman/model.py
hobarrera: It's a shame that click doesn't support RGB yet. Let's leave this as-is for now then.

hobarrera:
LGTM

untitaker:
I suspect we did *something* wrong, perhaps the escape sequence we tried out was malformed. I am not sure if that much has changed since this code works perfectly fine on xterm as well, and xterm doesn't change that much.

hobarrera:
*shurgs in ignorance*

At least this works now. I'll see if I can patch click to support rgb-colours.

PR #56, opened 2016-11-07
Add support for DTSTART field

hobarrera:
This field is commonly used as the reminder date on smartphone apps.


untitaker:
Can we have a default of `None`? I get notifications for every DTSTART in OpenTasks on Android.

PR #58, opened 2016-11-20
Add dtstart parameter, remove default

untitaker:


todoman/cli.py
```diff
@@ -52,13 +52,13 @@ def _todo_property_options(command):
         help=('The due date of the task, in the format specified in the '
               'configuration file.'))(command)
     click.option(
-        '--start', '-s', default='', callback=_validate_date_param,
+        '--dtstart', '-s', default='', callback=_validate_date_param,
```
hobarrera: I'd rather call this `start`. Even if vCard field name is `dtstart`, I think that for most users, start is more understandable (we *understand* vCard stuff, but most don't).

untitaker: I thought similarly, but dtstart (the Todo property) is exposed through the `--sort` option, so we would have to rename that as well.

hobarrera: I'd much rather expose `start` than `dtstart` (including in `sort`).

untitaker:
Renamed `Todo.dtstart` to `Todo.start`.

hobarrera:
LGTM. Feel free to merge this in! Thanks!

untitaker:
:100:

PR #59, opened 2016-11-24
Use TOML for the configuration file

### LINKED ISSUE #31: Preemptively validate configuration
While working on #30, I saw that todoman just passes the raw configparser through the CLI. I think it would be better if there was a dedicated Config type that has all valid params as properties (which raise if somebody tries to set an invalid value).

khal uses ConfigObj for this.

hobarrera:
Use TOML as a config format:
 * TOML is clearer than our current config format, unambiguous, and other pros I believe we already know.
 * It may be a bit premature though, since the current format hasn't given any problems.
 * TOML is becoming a lot more popular, especially in the python community.

Set default values when we load the config, rather than scatter them around the app, which should make everything a bit cleaner.

Closes #31

### DIFF
--- a/todoman/configuration.py
+++ b/todoman/configuration.py
@@ -1,11 +1,27 @@
-from configparser import ConfigParser
 from os import environ
 from os.path import exists, join
 
 import xdg.BaseDirectory
+import toml
 from click import ClickException
 
 from . import __documentation__
+
+
+defaults = {
+    'main': {
+        'color': 'auto',
+        'date_format': '%Y-%m-%d',
+    }
+}
+
+
+def merge_dicts(a, b):
+    for k, v in b.items():
+        if isinstance(v, dict) and k in a:
+            merge_dicts(a[k], v)
+        else:
+            a[k] = v
 
 
 def load_config():
@@ -19,7 +35,7 @@ def load_config():
         return _load_config_impl(custom_path)
 
     for d in xdg.BaseDirectory.xdg_config_dirs:
-        path = join(d, 'todoman', 'todoman.conf')
+        path = join(d, 'todoman', 'todoman.toml')
         if exists(path):
             return _load_config_impl(path)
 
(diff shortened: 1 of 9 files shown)


hobarrera:
@untitaker I'd love some feedback on whether this is worthwhile. Not sure it's of any actual use, though it makes sense for have todoman and vdirsyncer use the same format.

untitaker:
I am really not sure whether adding a new dependency is worth it if you currently have two config keys, both of which are strings.

Switch to TOML won't happen that soon for vdirsyncer. The intermediate goal in vdirsyncer's case is to get everybody to switch to quoted strings in their config.

todoman/configuration.py
```diff
@@ -31,6 +47,8 @@ def load_config():
 
 
 def _load_config_impl(path):
-    config = ConfigParser(interpolation=None)
-    config.read(path)
-    return config
+    with open(path) as conffile:
+        config = toml.loads(conffile.read())
+    # TODO: Validate required fields here (ie: path)
```
untitaker: If you're adding dependencies already you might as well use JSON-Schema for this

hobarrera: Huh, yeah, I'll look at validating the config file, since that *would* be a stronger bonus. But I don't see how to do this without another dependency (python-jsonschema).

hobarrera:
Added schema validation, and some tests. I believe we can also clean up a few other bits of code, since we can always assume that all configuration variables exist, and are valid.

untitaker:
Implementation looks good to me. This adds 0.1s overhead on my machine though.

untitaker:
Oh, I see that you're using `toml`, I'd rather recommend `pytoml`. Its code quality is better.

hobarrera:
> I'd rather recommend pytoml

I'll take a look at it. The interface seems pretty much the same.

untitaker:
Yeah the interface is always json-ish. But `toml.loads` is 230 lines.

hobarrera:
I've opted for configobj (#31) rather than toml.

* `ini` works well enough for us.
* Making users change their config file is a downside.
* Errors for pytoml are rather unfriendly (See avakar/pytoml#5)
* The validation schema syntax for JSON (which is what's used here) is horrible.

PR #61, opened 2016-12-30
Cache Todos with pickle

untitaker:
This is an optimization I've been playing around. It scrapes 0.05s from the
execution speed in my case (with warm cache), so I guess it does something
right. The improvement may be bigger on different datasets.

I wonder where to store the pickle cache, storing it in vdir seems unsafe.

### DIFF
--- a/todoman/model.py
+++ b/todoman/model.py
@@ -1,3 +1,4 @@
+import pickle
 import logging
 import os
 from datetime import date, datetime, time, timedelta
@@ -236,6 +237,7 @@ class Database:
 
     def __init__(self, path):
         self.path = path
+        self._cache_path = os.path.join(self.path, '.todoman-cache')
 
     @cached_property
     def todos(self):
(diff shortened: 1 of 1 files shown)


todoman/model.py
```diff
@@ -236,30 +237,52 @@ class Database:
 
     def __init__(self, path):
         self.path = path
+        self._cache_path = os.path.join(self.path, '.todoman-cache')
```
hobarrera: $XDG_DATA_HOME sounds okay for this:

https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html#variables

hobarrera:
> It scrapes 0.05s from the execution speed in my case (with warm cache)

I'm really not sure if the time difference warrant this. Do you have a very very small test vdir? I think that skipping non-VTODO entries by caching them as well might give a larger boots (since we might avoid opening dozens of irrelevant files).

untitaker:
>I think that skipping non-VTODO entries by caching them as well might give a larger boost (since we might avoid opening dozens of irrelevant files).

I tried that one as well. It doesn't have any effect. The conditional that checks whether `\nBEGIN:VTODO` is contained in the file helps massively though.

untitaker:
I fixed a bug where it didn't store entries for non-VTODO files.

untitaker:
Ok now it scrapes 0.12s from `time todo list`

todoman/model.py
untitaker: I opted for using `XDG_CACHE_HOME`

hobarrera:
Nice. Skipping non-VTODO files is a massive difference.

hobarrera:
LGTM, feel free to merge this.

PR #65, opened 2017-01-09
List authors and document copyright ownership

hobarrera:
Someone (I believe that it was @untitaker) mentioned that #5 was a bad idea since it ommited authors of copy-pasted code, and things like that.

@untitaker: I'd like your opinion on whether this is okay (or you want to chime in, in some way). It looks pretty much like what [vdirsyncer](https://github.com/pimutils/vdirsyncer) and [khal](https://github.com/pimutils/khal) are doing.

@geier I'm adding you, since I copied some code from khal, just letting you know, hope it's okay.

Closes #5

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -1,0 +1,16 @@
+Authors
+=======
+
+All contributors retain copyright to their contributions. Below is a list of
+authors who've contributed code to the project, or who wrote code elsewhere
+that we reused here.
+
+We manually track this since commit author *does not* equal code author.
+
+* Benjamin Frank <ben+gitlab@pipsfrank.de>
+* Christian Geier <geier@lostpackets.de>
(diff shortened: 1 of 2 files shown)


untitaker:
Lgtm

geier:
**I** am totally fine with this, but depending on which parts of khal you took, you may need/want to attribute some other people.

hobarrera:
Thanks for pointing that out, @geier. I basically copied all of `ExtendedEdit`. Inspecting the git logs reveals that @sithglan also authored a great deal of it, so it seems fair to list him as well. (@sithglan, any objections?).

Slightly OT: How do FLOSS projects usually deal with this? I mean, sometimes the logs are not so terribly clear and it's hard to pintpoint exact authors, what happens when you copy a chunk of code or things like that?

PR #66, opened 2017-01-11
Create interactive command for curses interface (for discussion purposes)

Pieter1024:
This is my first attempt at a curses-based interface for Todoman. I would like your feedback on what I've written so far. I am aware that it is by no means done, but before I start polishing it I would like to know I'm polishing the right thing.

To run:
* Create a virtualenv as you would for development purposes
* Run `todo interactive`

In the database listing view (the one you see when starting the program), the following commands are available:
* `l`: ListChooser, choose a different database from the available ones
* `e`: Edit, edit the currently highlighted item
* `n`: New, create a new item in this database
* `m`: Move, move the currently highlighted item to a database
* `c`: Copy, copy the currently highlighted item to a database
* `d`: Delete, delete the currently highlighted item
* `D`: Delete all completed, delete all tasks in the database that are completed
* `h`: Hide or show completed tasks in the database
* `j` and `k` are like 'up' and 'down', like in vim
* `space` or `enter` toggles the currently highlighted item

In all screens, `esc` closes the current screen and returns you to the previous one.

I have one specific issue on which I would appreciate your views. When I create a new item, the current database should be updated. However, you defined the todos property of Database as a cached_property, which prevents this. I have, for now, changed the cached_property into a property, but I have already noticed this is a severe performance impairment. Could we resolve this situation in a neater way? E.g., could we explicitly add the new item to the cached database?

The following definitely still needs to be done:

* Help screen
* Status bar messages
* Some configuration options (hide_completed, default_list, sort_order)

I would also like to make a page that shows all upcoming tasks across databases.


Pieter1024:
First bug found: if I create a new item, it defaults to completed, without regard to the information the user provided in the editor.

todoman/interactive.py
```diff
@@ -0,0 +1,583 @@
+import urwid
+from .ui import TodoEditor, EditState
+from .model import Todo
+
+import logging
+logging.basicConfig()
+
+logger = logging.getLogger()
+
+class TodomanItem(urwid.CheckBox):
```
hobarrera: `TodoItem` might be more descriptive.

As for other classes, prefixing `Todoman` to their name sounds redundant (especially since they're already in package `todoman.interactive`.

todoman/interactive.py
```diff
@@ -0,0 +1,583 @@
...
+    def __init__(self, filename, database, labelmaker):
+        '''
+        Create a TodomanItem instance based on the filename and the associated
+        Database that was provided. By providing the filename instead of the
+        Todo itself, we do not need to check if the provided todo is indeed in
+        in the given Database. The labelmaker is a function that turns a given
+        TodomanItem (i.e. self) into a string representation that is suitable
+        for its context.
+
+        (TodomanItem, str, Database, function) -> None
```
hobarrera: We're using sphinx-like doc for param/return types:

http://www.sphinx-doc.org/en/stable/domains.html#info-field-lists

It's nitpicky of me, but let's keep it consistent.

todoman/interactive.py
```diff
@@ -0,0 +1,583 @@
...
+    def keypress(self, size, key):
+        '''
+        Make the different commands in the TodomanItemListPage view work.
+
+        (TodomanItemListPage, int(?), str) -> str
+        '''
+        if key == 'l':
+            self.list_chooser()
+            return None
+        if key == 'm':
```
hobarrera: Use `elif` in all these and avoid all the extra `return None` lines.

todoman/interactive.py
```diff
@@ -0,0 +1,583 @@
...
+        self.loop.run()
+
+    def unhandled_input(self, key):
+        '''
+        Handles all the key presses that are application-wide.
+
+        (TodomanInteractive, str) -> None
+        '''
+        if key in ('q', 'Q'):
+            raise urwid.ExitMainLoop()
```
hobarrera: When editing an entry, if a checkbox if focus and I press <kbd>q</kbd>, it immediately exists, without any confirmation, etc.

todoman/model.py
```diff
@@ -237,7 +237,7 @@ class Database:
     def __init__(self, path):
         self.path = path
 
-    @cached_property
+    @property
```
hobarrera: You'll need to manually cache this to avoid the extra overhead.

That is:
* Set `_cached_todos` on first read.
* Manually add new entries there.
* Manually remove entries from there.
* Return `_cached_todos` if it's defined, otherwise cached all of them.

hobarrera:
Also, please rebase onto `master` and add yourself to AUTHORS.

todoman/model.py
untitaker: I think this was done to keep the database fresh even after editing an item. Let's make `todos` not a property, but instead rename it to `get_todos` with a `skip_cache` parameter.

untitaker:
I'd prefer if the interactive interface does deletions and marking things as done upon exiting the interface, so I can hit `^C` to discard those changes.

todoman/cli.py
```diff
@@ -179,6 +184,14 @@ def edit(ctx, id, todo_properties, interactive):
...
         ctx.exit(1)
 
+@cli.command()
+@click.pass_context
+def interactive(ctx):
+    '''
+    Provide an interactive, curses-based interface to Todoman.
+    '''
+    TodomanInteractive(ctx.obj['db'].values(), ctx.obj['formatter'])
+    ctx.exit(1)
```
untitaker: Why always exit with 1?

todoman/interactive.py
```diff
@@ -0,0 +1,583 @@
...
+        '''
+        self.footer.contents
+        self.footer.contents[0][0].original_widget.set_text(text)
+
+    def callback(self, **kwargs):
+        '''
+        A default callback function to use when closing the previous page. 
+        This callback function handles several keywords that are generic to
+        all callbacks. If a certain keyword is not set, the method does
+        nothing with it.
```
untitaker: I think pages could be implemented *much* more easily, without subclassing. You'd need a wrapper widget that catches `Esc` (so just let the keypress event bubble up from the page's handler in that case), and that widget maintains a stack (list) of pages that it appends and pops from. If the list is empty, `ExitMainLoop` is raised.

todoman/interactive.py
```diff
@@ -0,0 +1,583 @@
...
+        will be opened over the old one.
+
+        This callback function handles the following keywords, in addition
+        to the keywords TodomanPage.callback handles:
+
+        * database (Database): the Database to display in the new page.
+
+        (TodomanItemListPage, **kwargs) -> None
+        '''
+        for key, value in kwargs.items():
```
untitaker: Same here, why not just have `database` as part of the method signature?

todoman/interactive.py
```diff
@@ -0,0 +1,583 @@
...
+        Todo itself, we do not need to check if the provided todo is indeed in
+        in the given Database. The labelmaker is a function that turns a given
+        TodomanItem (i.e. self) into a string representation that is suitable
+        for its context.
+
+        (TodomanItem, str, Database, function) -> None
+        '''
+        self.filename = filename
+        self.database = database
+        try:
```
untitaker: Might as well have the params `database, todo` instead of `database, filename` and avoid this logic.

todoman/interactive.py
```diff
@@ -0,0 +1,583 @@
...
+        '''
+        Create a list of TodomanItems to display, based on the associated
+        Database and the current done_is_hidden setting.
+
+        (TodomanItemListPage) -> [TodomanItem]
+        '''
+        items = []
+        for filename in self.database.todos.keys(): 
+            todo = TodomanItem(filename, self.database, self.generate_label)
+            if not self.done_is_hidden or not todo.is_completed:
```
untitaker: We should reuse the logic from the CLI here, but not sure how.

todoman/interactive.py
```diff
@@ -0,0 +1,583 @@
...
+
+    def items_to_display(self):
+        '''
+        Create a list of TodomanItems to display, based on the associated
+        Database and the current done_is_hidden setting.
+
+        (TodomanItemListPage) -> [TodomanItem]
+        '''
+        items = []
+        for filename in self.database.todos.keys():
```
untitaker: If `TodomanItem` accepts a `todo` instead of a `filename`, you can just iterate through values here and save one key lookup in `TodomanItem.__init__`

todoman/interactive.py
```diff
@@ -0,0 +1,583 @@
...
+        items.sort(key = lambda item: item.label.lower())
+        return items
+
+    def callback_move_to(self, **kwargs):
+        '''
+        Move the TodomanItem in focus to Database database.
+
+        (TodomanItemListPage, Database) -> None
+        '''
+        for key, value in kwargs.items():
```
untitaker: Same here, just `kwargs.pop` or put `database` into the method signature.

todoman/interactive.py
```diff
@@ -0,0 +1,583 @@
...
+        '''
+        Open a TodomanDatabasesPage from which to choose the destination of the
+        copy operation for the selected item.
+
+        (TodomanItemListPage) -> None
+        '''
+        newPage = TodomanDatabasesPage(self.parent, self.callback_copy_to)
+        self.open_page(newPage)
+
+    def delete(self, item = None):
```
untitaker: Please no spaces between `=` when part of parameters.

todoman/interactive.py
```diff
@@ -0,0 +1,583 @@
...
+        self.body.body.remove(item)
+
+    def generate_label(self, item):
+        '''
+        Return a label for a given TodomanItem for display in the
+        TodomanItemListPage listing.
+
+        (TodomanItemListPage, TodomanItem) -> str
+        '''
+        return "{0} {1}".format('!' if item.has_priority else ' ',
```
untitaker: We already have `todoman.ui.TodoFormatter` for this (which will also add colors)

todoman/cli.py
```diff
@@ -9,6 +9,11 @@
 from .main import dump_idfile, get_task_sort_function, get_todo, get_todos
 from .model import Database, Todo
 from .ui import EditState, TodoEditor, TodoFormatter
+from .interactive import TodomanInteractive
+
+import logging
+logging.basicConfig()
```
untitaker: `basicConfig` should be called in the `cli` function, not at module level. In the case of showing a UI you won't want to output anything I think.

todoman/cli.py
```diff
@@ -9,6 +9,11 @@
...
+logger = logging.getLogger()
```
untitaker: Use `logging.getLogger(__name__)` to namespace your logging.

todoman/interactive.py
```diff
@@ -0,0 +1,583 @@
...
+        all callbacks. If a certain keyword is not set, the method does
+        nothing with it.
+
+        This callback function supports the following keywords:
+
+        statusbar (str): A text to set as the statusbar message
+
+        (TodomanPage, **kwargs) -> None
+        '''
+        for key, value in kwargs.items():
```
untitaker: Isn't it invalid to call this callback with anything but `statusbar`?

* If yes: `def callback(self, statusbar)`
* If no: `def callback(self, statusbar, **kwargs)`

untitaker:
I generally feel like a lot of classes are unnecessarily prefixed by `Todoman`.

untitaker:
@Pieter1024 it's been a while. Are you still willing to work on this?

hobarrera:
Sorry, I really broke this PR with some substantial changes to `todoman`.  
I'll can rebase it into master if you're still interesting on working on it, @Pieter1024.

hobarrera:
I've rebased this to master and fixed style violations, but haven't really looked at the code more closely. Since @Pieter1024 hasn't responded back, will probably be picking this up in the coming days, if time allows.

untitaker:
I'll make a separate PR for the logging.

untitaker:
See #106

todoman/interactive.py
```diff
@@ -530,7 +530,8 @@ def __init__(self, databases, formatter):
         Create a Main instance based on the Database objects that
         the regular Todoman cli module passes.
 
-        (Main, [Database], TodoFormatter) -> None
+        :type databases: list[Database]
```
hobarrera: We still need to change this, because with the current model it doesn't make sense to pass a list of `Database`s.

todoman/interactive.py
```diff
@@ -62,10 +60,8 @@ def has_priority(self):
         '''
         Returns True iff the Item refers to a Todo with a priority that
         is set (i.e. not None and above zero).
-
-        (self) -> bool
         '''
-        return self.todo.priority not in [None, 0]
+        return self.todo.priority not in (None, 0)
```
hobarrera: Maybe it's time to implement `Todo.is_urgent` as a wrapper around this, so we stop duplicating `todo.priority not in [None, 0]` al over the place.

Also, I think it can never be None, because of how the getter is written.

todoman/interactive.py
```diff
@@ -167,7 +156,9 @@ def __init__(self, parent, callback, database):
         is responsible for passing information to the underlying Page.
         It is usually a method of the underlying Page.
 
-        (ItemListPage, Main, function, Database) -> None
+        :param parent: Main
```
hobarrera: That's not the right format, it's `:param TYPE NAME`, for example `:param int amount`.

hobarrera:
Thanks. I also think we should move packages around:

* `ui`: has some interactive stuff, and formatters.
* `interactive`: has more interactive stuff.

I think this should end up in:

* `ui`/`interactive` (don't care which name): all the interactive stuff.
* `formatting`: Just the formatters (shared between cli and interactive).

todoman/interactive.py
untitaker: We can remove this if we use the normal formatter instead of `generate_label`

untitaker:
Move `TodoEditor` into this module, rename `ui` to `formatter` and move `cli, formatter, interactive` into a new `ui` subpackag

untitaker:
I just tried to refactor and bumped into a few design limits of the current datamodel

hobarrera:
Proposal seems great.

Do these limitations block the refactor? Or are they for future reference?

untitaker:
I think so, the API isn't very pleasing for this usage.

Particularly the split between `Todo` and `FileTodo` is problematic.

hobarrera:
Maybe I'll have to revisit that. Having a separate class keeps the door open for separate storage types (will we ever want that!?). And I think is a good abstraction.

Maybe having the `save` logic in the database is a better choice (as far as abstraction goes), but we need to track "read from file" vs "read from cache". Maybe a flag in `Todo` will do (`readonly`?).

Issue #23, opened 2016-04-18
Implement an SQLite cache


hobarrera:
Use a cache for:
- Faster startup times
- Keeping a proper index of `id`s to show
- Simplify filtering/sorting by merely using SQL

hobarrera:
I'm currently working on this, and re-writing the whole model backend, so I'd suggest not doing intensive works in those areas in the meantime.

untitaker:
I think it's a bad idea to expose SQL queries.

hobarrera:
> I think it's a bad idea to expose SQL queries.

They will only be used internally. The sorting command and all the cli commands in general will remain the same.

hobarrera:
I had to dump all this, since I'd not considered recurring events. The rewrite will have to take this into consideration.

untitaker:
We now have two caches:

* A JSON object that maps ids to filenames
* A pickled dict that caches all `.ics` files

Can we close this issue?

hobarrera:
The idea behind this is to cache all these fields in a small sqlite db, and simply use SQL to filter (not exposing this to the user, of course).  This makes the code on #68 a lot friendlier.

Since we only use these fields in `list()`, we don't even need to read files themselves (except out-of-date cache) except when edition/viewing a single `todo`.

Also, it means that we can keep `id`s valid for much longer times (since we can persist them there too). I'd eventually re-compute them on `flush()`.

untitaker:
Persisting IDs may be a bit annoying because those will eventually become larger, and therefore be harder to type.

PR #68, opened 2017-01-12
Implement the `remind` command

### LINKED ISSUE #67: List by "today", "this week"
When I use `todo list` I get a very long list of all my tasks. I know I can filter by lists, but is it possible to filter by "this week", "today", "overdue"? I'd like to have a short list of today's (and overdue) tasks so I get a good feeling when I've done all for today :-)

hobarrera:
The main idea behind this is to allow external scripts to remind of soon-to-expire tasks, or just allow the user to check if they're up to date on tasks or now.

Some stuff that needs to be defined:

 * Do we want `due_in` to be a flag or configuration setting? I'm inclined to make it the latter, though both might be possible.
 * Do we want `overdue` to be a flag or configuration setting? Same doubt here.

Pending:

* [x] Tests:
  * Include ones with tz-aware tasks, and tz-naive tasks.
* [ ] A `--porcelain` flag, to keep the format permanently frozen so scripts can parse this easily.

I think it's safe to say that this would close #67

Please, chime in on the functionality (I'll continue polishing the code in the meantime, it's a mere prototype)

### DIFF
--- a/todoman/cli.py
+++ b/todoman/cli.py
@@ -304,7 +304,13 @@ def move(ctx, list, ids):
 @click.option('--reverse/--no-reverse', default=True,
               help='Sort tasks in reverse order (see --sort). '
               'Defaults to true.')
-def list(ctx, lists, all, urgent, location, category, grep, sort, reverse):
+@click.option('--due', default=None, help='Only show tasks due in DUE hours',
+              type=int)
+# TODO: we might want a `porcelain` flag here to print this is a
+# machine-friendly format that NEVER CHANGES!
+def list(
+    ctx, lists, all, urgent, location, category, grep, sort, reverse, due,
+         ):
     """
     List unfinished tasks.
 
@@ -323,6 +329,7 @@ def list(ctx, lists, all, urgent, location, category, grep, sort, reverse):
 
     db = ctx.obj['db']
     todos = db.todos(
+        due=due,
         all=all,
         category=category,
         grep=grep,
--- a/tests/test_filtering.py
+++ b/tests/test_filtering.py
@@ -1,4 +1,7 @@
+from datetime import datetime, timedelta
+
 from todoman.cli import cli
+from todoman.model import Database
 
 
 def test_all(tmpdir, runner, create):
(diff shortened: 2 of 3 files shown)


untitaker:
I don't think this has to be its own command, but rather a set of new flags on
`list` (tend to be against config settings). Particularly the porcelain flag
sounds nice to have even for other usage of `list`.

hobarrera:
Good call. It's also *a lot* less code than I'd expect. It's a shame that the filtering `for` loop keeps getting uglier, but I have come plans in mind for that one. ^^

todoman/cli.py
```diff
@@ -294,7 +296,13 @@ def move(ctx, list, ids):
 @click.option('--reverse/--no-reverse', default=True,
               help='Sort tasks in reverse order (see --sort). '
               'Defaults to true.')
-def list(ctx, lists, all, urgent, location, category, grep, sort, reverse):
+@click.option('--due', default=None, help='Only show tasks due in DUE hours',
+              type=int)
+# TODO: we might want a `porcelain` flag here to print this is a
```
untitaker: JSON seems like the obvious candidate. There are CLI tools dedicated to selecting fields from JSON, for scripting purposes.

todoman/cli.py
```diff
@@ -326,11 +338,23 @@ def list(ctx, lists, all, urgent, location, category, grep, sort, reverse):
                (not pattern or (
                    pattern.search(todo.summary) or
                    pattern.search(todo.description)
-                   ))
+                )) and
+               (
```
untitaker: On all other lines the following style is used:

    (a
     b)

but here you use:

    (
      a
      b
    )

todoman/cli.py
hobarrera: I was thinking something a bit simpler, like this (due date, percent, title). Basically inspired in `git status --porcelain`:

2017-01-01 0 Do stuff
2017-01-01 100 Do more stuff

---

JSON is nicer for apps to read, but a bit more complex for bash scripts. I guess it depends a bit on who we're targeting with this, but I'd much rather stick to the simplest-possible format.

untitaker: For scripting purposes https://github.com/stedolan/jq can be used

todoman/cli.py
```diff
@@ -325,12 +337,22 @@ def list(ctx, lists, all, urgent, location, category, grep, sort, reverse):
...
-                   pattern.search(todo.description)
-                   ))
+                   pattern.search(todo.description))) and
+               (not due or
+                (todo.due and
+                 not todo.due.tzinfo and
+                 (todo.due <= max_naive)) or
+                (todo.due and
+                 todo.due.tzinfo and
+                 (todo.due <= max_aware)))
```
untitaker: This is getting out of hand, I think it's time for a new function

hobarrera: Yes, I kind of makes me lose faith a bit and just work on #23.

untitaker: Moving this into a function is really easy compared to implementing a new cache.

todoman/cli.py
hobarrera: You've convinced me; +1 for JSON.

todoman/cli.py
hobarrera: IIRC, python does some magic optimizations for this sort of list comprehension, I'd have to see if they still apply here, and if a function has any drawbacks (I admit I'm a bit ignorant on this, and might be mistaken).

untitaker: I'd argue that if you care about this, you'd have to rewrite it in a compiled language.

PR #72, opened 2017-01-26
Implement an SQL cache

### LINKED ISSUE #23: Implement an SQLite cache

hobarrera:
I've had this in mind since like forever, with some stashed bits here and there...

**Pros**
* Faster: For my local data, current implementation runs `list` in ~~(890ms/770ms (without/with an existing cache). This runs at 440ms/365ms. Numbers may vary, but should be proportional.)~~ This has improved further. Time reduction is ~77% for me.
* Simpler querying interface: We can write SQL to filter stuff out, making it a lot simpler.
* ~~(Higher test coverge (even though there are no new tests, it's just how the code is written).)~~ Nope, there's just more code.
* Just one cache mechanism.

**Cons**
* The code might look a bit more complex until you get a grasp of it.
* I'm pretty sure I had a very clear one that I'm forgetting right now.

**Subjective**

* `id`s are persistent now, which makes some things a bit simpler/safer (eg: they won't change between runs, unless flushed, or the cache is purged).

**Pending**
* [x] `move` test is failing. No idea why, yet.
* [x] `copy` test is failing. No idea why, yet.
* [x] `delete` test is failing. No idea why, yet.
* [x] `flush` should entirely delete the cache (or maybe another command?). This also resets `id`s.
* [x] cache location should be configurable.
* [x] revise `Todo` flags `new` and `safe`.
* [x] check `list LISTNAME` fails. (there are no tests for that yet).
- [x] revise usage of cursor (@untitaker)

**Pending?**
* ~~(I might rip the `Cache`/`Database` distinction and shape the whole thing into a single class.)~~
---

Yes, I *will* squash these commits. Feel free to give any feedback. Hopefully this isn't too long and painful to have a look at.

Closes #23


todoman/model.py
```diff
@@ -70,7 +87,21 @@ def __init__(self, todo=None, filename=None):
         if self.todo.get('dtstamp', None) is None:
             self.todo.add('dtstamp', datetime.utcnow())
 
+        self.safe = safe or new
         self.filename = filename or "{}.ics".format(self.todo.get('uid'))
+        self.mtime = mtime or datetime.now()
+
+    @staticmethod
```
untitaker: In principle this should be a `classmethod` such that this returns the correct subclass when `Todo` is subclassed.

todoman/model.py
```diff
@@ -47,7 +61,8 @@ class Todo:
 
     _localtimezone = tzlocal()
 
-    def __init__(self, todo=None, filename=None):
+    def __init__(self, todo=None, filename=None, mtime=None, safe=False,
```
untitaker: I'm unsure if I'm correctly understanding the context, but I think you meant to write `save` instead of `safe`.

todoman/model.py
```diff
@@ -231,74 +274,493 @@ def _normalize_datetime(self, x):
...
+            order = ','.join(order)
+        else:
+            order = '''
+                completed_at DESC,
+                priority ASC,
+                due IS NOT NULL, due DESC,
+                created_at ASC
+            '''
+
+        if not reverse:
```
untitaker: If `not reverse` is the more common codepath it may make sense to generate the "not-reverse" query by default, and apply a text-replacement only if `reverse=True`.

todoman/model.py
```diff
@@ -231,74 +274,493 @@ def _normalize_datetime(self, x):
...
+        self.cur.execute("PRAGMA foreign_keys = ON")
+
+        self.create_tables()
+
+    def load_from_disk(self):
+        conn_disk = sqlite3.connect(
+            self.cache_path,
+            detect_types=sqlite3.PARSE_DECLTYPES | sqlite3.PARSE_COLNAMES,
+        )
+        sqlitebck.copy(conn_disk, self.conn)
```
untitaker: I really don't understand why you need to copy databases around. Transactions should solve this (so a single `commit` after you're done).

untitaker: https://github.com/pimutils/todoman/pull/73

untitaker:
>I might rip the Cache/Database distinction and shape the whole thing into a single class.

-1, I think it's a good idea to keep them separate. It doesn't make sense for `move` to live on `Cache`. Also in the future this may make it easier to implement a different cache and hide it behind a feature flag.

todoman/cli.py
```diff
@@ -290,6 +302,7 @@ def move(ctx, list, ids):
 @click.option('--location', help='Only show tasks with location containg TEXT')
 @click.option('--category', help='Only show tasks with category containg TEXT')
 @click.option('--grep', help='Only show tasks with message containg TEXT')
+# XXX: shouldn't this be an ngarg-1 this?:
```
hobarrera: I mean: shouldn't `--sort` be a [multi value option](http://click.pocoo.org/6/options/#multi-value-options)? We actually split the string manually ourselves below.

untitaker: Multi-value options lose order of the items IIRC, but we do care about that.

I suspect even if Click preserves order that it wasn't intended by Armin

todoman/model.py
untitaker: Eh, nevermind. I don't understand what `safe` does

tests/conftest.py
```diff
@@ -9,7 +10,10 @@
 
 @pytest.fixture
 def default_database(tmpdir):
-    return model.Database(path=str(tmpdir.mkdir('default')))
+    return model.Database(
+        [str(tmpdir.mkdir('default'))],
+        os.path.join(str(tmpdir.mkdir(uuid4().hex)), 'cache.sqlite3'),
```
hobarrera: Nice, thanks.

tests/test_basic.py
```diff
@@ -160,8 +160,8 @@ def test_dtstamp(tmpdir, runner, create):
     result = runner.invoke(cli, ['new', '-l', 'default', 'test event'])
     assert not result.exception
 
-    db = Database(str(tmpdir + '/default'))
-    todo = list(db.todos.values())[0]
+    db = Database([str(tmpdir + '/default')], str(tmpdir + '/dtstamp_cache'))
```
hobarrera: It's more readable, and I don't mind the refactor now, since it doesn't add noise to the project log anyway (it gets squashed).

todoman/model.py
hobarrera: Updated docstring so this is clearer.

todoman/model.py
hobarrera: Interesting point. Hadn't thought of that. Fixed.

todoman/model.py
hobarrera: `reverse` is `True` by default when running via cli. I've updated this method's default to be the same.

todoman/model.py
hobarrera: I was doing this for speed - to reduce disk I/O. But it seems that my initial testing was mistaken and this just made the whole thing slower. (time on my same test case without this is about 1/3 of the original one!).

untitaker:
AFAICT the remaining test failures are because you completely forgot to validate whether files in the cache still exist on disk.

todoman/model.py
```diff
@@ -231,74 +276,472 @@ def _normalize_datetime(self, x):
...
+        extra_where = []
+        params = []
+
+        if not all:
+            extra_where.append('AND completed_at is NULL')
+        if lists:
+            q = ', '.join(['?'] * len(lists))
+
+            query = 'SELECT id FROM lists WHERE name IN ({});'.format(q)
+            list_ids = [row['id'] for row in self.conn.execute(q, lists)]
```
untitaker: Given this, wouldn't it make sense to make `list.name` the primary key? (not sure if possible in sqlite)

todoman/model.py
```diff
@@ -231,74 +283,467 @@ def _normalize_datetime(self, x):
...
+
+        extra_where = []
+        params = []
+
+        if not all:
+            extra_where.append('AND completed_at is NULL')
+        if lists:
+            q = ', '.join(['?'] * len(lists))
+
+            query = 'SELECT id FROM lists WHERE name IN ({});'.format(q)
```
untitaker: The value assigned here doesn't seem to be used at all.

hobarrera: The query on 473 doesn't work and I've no idea why (also, it needs a unit test to cover it).

untitaker:
I have a fix for this, hang on

todoman/model.py
hobarrera: Not only does it work, it also seems to slightly increase speed.

todoman/model.py
```diff
@@ -456,6 +456,7 @@ def add_todo(self, todo, file_id):
 
     def todos(self, all=False, lists=[], urgent=False, location='',
               category='', grep='', sort=[], reverse=True):
+        lists = [l.name if isinstance(l, Todo) else l for l in lists]
```
hobarrera: I don't understand this line, and it makes most tests fail.

untitaker:
Sorry, I broke something! Give me a few minutes!

todoman/model.py
hobarrera: Oh, you immediately fixed it. ^^

hobarrera:
> Sorry, I broke something! Give me a few minutes!

np

untitaker:
In fact it appears that https://github.com/pimutils/todoman/pull/72/commits/198ead844dedacd207a174efd20a518d4ea99d77 broke something, not me.

hobarrera:
> In fact it appears that 198ead8 broke something, not me.

Oh, I might have screwed up rebasing. In any case, it's all working now (except for move/delete).

untitaker:
It doesn't appear so to me:

    E       sqlite3.OperationalError: no such column: files.list_name

At `model.py:514`

hobarrera:
If running non-tests, make sure you delete your old database cache::

    rm ~/.cache/todoman/cache.sqlite3

hobarrera:
I'll add versioning later on (probably when I make some change to the schema).

untitaker:
In fact I had to remove the tests to run the tests successfully! Something is leaking here.

hobarrera:
> In fact I had to remove the tests to run the tests successfully! Something is leaking here.

The cache used during tests is the one in your home. :(  
I'll have to make cache_path configurable and override that for tests.

untitaker:
Yes please, this will make tests much faster.

untitaker:
Is there a reason why files and todos are in separate tables?

hobarrera:
A file might have a calendar entry and no todo. By keeping this cached, we avoid ever re  reading.

untitaker:
I refactored everything and now the tests pass

untitaker:
I'm not particularly happy about todo IDs being persistent. I have many unfinished tasks that are hidden away in other lists, and their existence bumps the IDs of the tasks that I do currently focus on.

todoman/model.py
```diff
@@ -602,34 +609,22 @@ def todo(self, id):
...
         """
-        row = self.conn.execute(
-            "SELECT mtime FROM files WHERE path = ?",
-            (path,)
-        ).fetchone()
-        if row is None:
-            return False
-        if int(row['mtime']) == mtime:
-            return True
+        for row in self.cur.execute("SELECT path, mtime FROM files"):
```
hobarrera: I expected this to be slow, but wow, it's fast. I'm impressed! :+1:

todoman/model.py
```diff
@@ -628,21 +620,14 @@ def expire_files(self, paths_to_mtime):
...
     def _expire_file(self, path):
-        row = self.cur.execute(
-            "SELECT id FROM files WHERE path = ?",
-            (path,),
-        ).fetchone()
-
-        assert row is not None
-        file_id = row['id']
-
+        # XXX: Isn't this covered by the second cmd?
```
hobarrera: I don't understand this comment.

hobarrera:
@untitaker I'd like to already squash these commits and prepare this PR for merging. Anything further to add? (importantly: anything you haven't pushed that would be painful to merge if I squash this?).

todoman/model.py
untitaker: `ON DELETE CASCADE` should delete the respective todo if we delete the file.

untitaker:
Two things:

* Not sure if adding the `list` property to todos is a good idea
* The speedups are nowhere near what you're seeing. I'm getting 100ms improvements at most

hobarrera:
> ON DELETE CASCADE should delete the respective todo if we delete the file.

Yes, it does. I think that code is from before I realized I had to manually enforce FKs.

> Not sure if adding the list property to todos is a good idea.

Why?

> The speedups are nowhere near what you're seeing. I'm getting 100ms improvements at most

I just checked this more: I get these speedups in a virtualenv, but not when installing properly. Not sure what's going on.

untitaker:
Particularly overriding the `list` param in the `copy` cmd looks like it might introduce mutability bugs one day.

I don't really see the need for Todo to know about Lists.

hobarrera:
`Todo`s use their `List` to construct the absolute file path. I actually thought it might help make a "list-picker" for the edit UI at some point. But I'll accept alternatives to it though.

untitaker:
Re listpicker: I'd save that refactor for when we actually need it.

hobarrera:
> I had this thought too, but I didn't consider it to be a good API that expire_files would mutate its parameters. In any case the old code didn't reparse anything.

Sorry, you're right, the `continue` would have made it skip, and the failed-`INSERT` doesn't really add anything. I'll revert that bit.

hobarrera:
I guess we could pass it when saving - ie: `def save(self, list_)`.  
My initial though is that `Database.todo()` could return a todo and its list as a tuple, and then pass that around when formatting, saving, etc.

The second option would be to rewrite `copy` and avoid mutating the original `Todo`. That sounds simpler and cleaner.

untitaker:
both options sound fine to me

todoman/cli.py
hobarrera: As a follow-up: I removed this. `option` can't have `nargs=-1`, only `argument` does.

hobarrera:
I'm ready to merge this, unless there' additional feedback.

untitaker:
Fine by me!

On 30 January 2017 08:17:25 CET, Hugo Osvaldo Barrera <notifications@github.com> wrote:
>I'm ready to merge this, unless there' additional feedback.
>
>-- 
>https://github.com/pimutils/todoman/pull/72#issuecomment-275994100

-- 

PR #76, opened 2017-01-30
Add tests to verify that we handle TZ properly

hobarrera:
I wrote this trying to track down what was wrong with #75, and now it's a free regression test (that didn't help).

### DIFF
--- a/tests/test_model.py
+++ b/tests/test_model.py
@@ -1,3 +1,7 @@
+from datetime import datetime
+
+from dateutil.tz.tz import tzoffset
+
 from todoman.model import Database
 
 
@@ -19,3 +23,27 @@ def test_querying(create, tmpdir):
     assert len(set(db.todos())) == 9
     assert len(set(db.todos(lists='ab'))) == 6
     assert len(set(db.todos(lists='ab', location='a'))) == 2
+
+
+def test_retain_tz(tmpdir, runner, create, default_database):
+    create(
+        'ar.ics',
+        'SUMMARY:blah.ar\n'
+        'DUE;VALUE=DATE-TIME;TZID=HST:20160102T000000\n'
+    )
+    create(
+        'de.ics',
+        'SUMMARY:blah.de\n'
+        'DUE;VALUE=DATE-TIME;TZID=CET:20160102T000000\n'
+    )
+
+    db = Database([tmpdir.join('default')], tmpdir.join('cache.sqlite'))
+    todos = list(db.todos())
+
+    assert len(todos) == 2
+    assert todos[0].due == datetime(
+        2016, 1, 2, 0, 0, tzinfo=tzoffset(None, -36000)
+    )
+    assert todos[1].due == datetime(
+        2016, 1, 2, 0, 0, tzinfo=tzoffset(None, 3600)
+    )


hobarrera:
Looks like `icalendar` might be doing something funky, and the test runs for me by coincidence (because ART is my local timezone, and is used for TZ-naive datetimes):

For `DUE;VALUE=DATE-TIME;TZID=CET:20160102T000000`
```
>>> todo.__class__
<class 'icalendar.cal.Todo'>
>>> todo.decoded('due')
datetime.datetime(2016, 1, 2, 0, 0, tzinfo=<DstTzInfo 'CET' CET+1:00:00 STD>)
```

For `DUE;VALUE=DATE-TIME;TZID=ART:20160102T000000`
```
>>> todo.__class__
<class 'icalendar.cal.Todo'>
>>> todo.decoded('due')
datetime.datetime(2016, 1, 2, 0, 0)
```

@untitaker Any insight on this, maybe from dealing from something similar in khal?

untitaker:
Your test confuses me. Why take an index from todos when you're already iterating through them?

I'll play around with this tomorrow.

tests/test_model.py
```diff
@@ -19,3 +23,27 @@ def test_querying(create, tmpdir):
     assert len(set(db.todos())) == 9
     assert len(set(db.todos(lists='ab'))) == 6
     assert len(set(db.todos(lists='ab', location='a'))) == 2
+
+
+def test_retain_tz(tmpdir, runner, create, default_database):
+    create(
+        'ar.ics',
+        'SUMMARY:blah.ar\n'
+        'DUE;VALUE=DATE-TIME;TZID=ART:20160102T000000\n'
```
untitaker: Doesn't the VTODO have to contain a `VTIMEZONE` component?

IIRC khal has more complex code for dealing with timezones. cc @geier

hobarrera: I'm not sure, since the other test case (`CET`) works fine.

geier: I'm not to familiar with todoman's code. What are you trying to do? (perhaps irc?)

hobarrera: The [comment above](https://github.com/pimutils/todoman/pull/76#issuecomment-276219104) has the relevant data isolated from todoman.

geier: ah, the issue is that icalendar checks if the TZID is in pytz.all_timezones ('CET' is, 'ART' isn't), and if it isn't uses one from the ics file, if there is one.

geier: todo.params probably contains a `TZID` parameter

untitaker: According to https://github.com/newvem/pytz/blob/master/pytz/zone.tab it may be "AR".

hobarrera: @geier Thanks. That pretty much answers everything. Out of curiosity: is this a "standard" list, or pytz-specific?

@untitaker AR isn't mentioned in `all_timezones` either (and CET isn't in the list on that link).

geier: 'America/Argentina/Buenos_Aires' should work, does Argentina really have that many different timezones today? I count 13, in the database, or is it just for historical purposes?

geier: @hobarrera pytz is based on the time-zones database, aka Olson database and therefore uses their identifiers, see http://www.iana.org/time-zones

geier: perhaps better: https://en.wikipedia.org/wiki/Tz_database#Names_of_time_zones

hobarrera: @geier Nope, I just used my own because I originally copied the line from some real todo.
Argentina is pretty disorganized when it comes to timezones; we have just one (UTC-3).

We don't have DST, *except* some years when a presidential order added a one-time DST, and provinces could individually adhere to it (or not). Since some provinces adhered to DST, we then had different timezones.

Since different provinces adhered to the order on different one-time DST-additions, this meant that they basically spawned their own timezone.

So, as far as IT does, we have potentially 23 timezones, which *are usually* the same, but will vary every several years.

To make matters worse, this all tends to be done a month or two before DST hits.

hobarrera: I ended up using "HST" though, since I like how the three-letter timezones look, and how the code aligns.

geier: sounds interesting...

Issue #69, opened 2017-01-13
Prevent adding a due date automatically


pfertyk:
I just started using todoman. I noticed that when I create a new task, it has a due date set automatically to now + 24h. Is there an option to disable this feature (other than editing each task and removing the due date manually)?

hobarrera:
This isn't currently supported. I guess we could have a setting `default_due` (in hours, with a default value of 24), which can supress this if the value is 0.

I'm open for PRs for this, can't quite pick it up right now.

pfertyk:
Took me some time, but I think I got it :) The PR has been created, could you please review it?

hobarrera:
Closed via #78

PR #78, opened 2017-02-01
Create configuration for default_due

pfertyk:
As mentioned in #69, this PR should allow for turning off the default 24h due date for new todos. The default_due will still be 24h if not specified. If set to 0, new todo's will not have the due date set automatically.


tests/test_basic.py
```diff
@@ -192,6 +192,42 @@ def test_default_list(tmpdir, runner, create):
     assert todo.summary == 'test default list'
 
 
+def test_default_due_is_not_specified(tmpdir, runner, create):
```
untitaker: You can collapse all these tests into one using http://doc.pytest.org/en/latest/example/parametrize.html

todoman/model.py
```diff
@@ -91,9 +93,12 @@ def __init__(self, todo=None, filename=None, mtime=None, new=False):
 
         if new:
             now = datetime.now(self._localtimezone)
+            config = load_config()
```
untitaker: This is not a good place to do this, particularly because the config is already read from disk, and calling `load_config` again parses it again.

I suspect that the `Todo` class just shouldn't set a `due` date, and that the entire logic belongs into the `cli` module (particularly into the `new` command).

untitaker: Also this breaks tests where the `Todo` class is tested without a config environment around it.

tests/test_basic.py
pfertyk: I think that would require at least 2 `if` statements (one for special case where no `default_due` is specified, one for special case where result is `None` if `default_due` is 0). That would make the tests less readable. Do you think it's worth the effort?

pfertyk: I'm working on it to see how those two `if` statements will look.

pfertyk: OK, it doesn't look that bad. The tests are now parametrized.

todoman/cli.py
```diff
@@ -148,6 +149,10 @@ def new(ctx, summary, list, todo_properties, interactive):
     if not todo.summary:
         raise click.UsageError('No SUMMARY specified')
 
+    default_due = int(ctx.obj['config']['main'].get('default_due', '24'))
+    if default_due:
+        todo.due = todo.created_at + timedelta(minutes=default_due*60)
```
untitaker: The styleguide mandates that there's spacing around the `*` (see PEP-8)

AUTHORS.rst
```diff
@@ -14,3 +14,4 @@ We manually track this since commit author *does not* equal code author.
 * José Ribeiro <hello@jlbribeiro.com>
 * Markus Unterwaditzer <markus@unterwaditzer.net>
 * Thomas Glanzmann <thomas@glanzmann.de>
+* Paweł Fertyk <pfertyk@openmailbox.org>
```
hobarrera: These entries should remain in alphabetical order.

tests/test_basic.py
```diff
@@ -192,6 +192,32 @@ def test_default_list(tmpdir, runner, create):
...
+
+    runner.invoke(cli, ['new', '-l', 'default', 'aaa'])
+    db = Database([tmpdir.join('default')], tmpdir.join('/default_list'))
+    todo = list(db.todos())[0]
+
+    if expected_due_hours is None:
+        assert todo.due is None
+    else:
+        assert (todo.due - todo.created_at) == datetime.timedelta(
+            minutes=expected_due_hours * 60
```
hobarrera: I think you can do `hours=expected_due_hours` here.

todoman/cli.py
```diff
@@ -148,6 +149,10 @@ def new(ctx, summary, list, todo_properties, interactive):
     if not todo.summary:
         raise click.UsageError('No SUMMARY specified')
 
+    default_due = int(ctx.obj['config']['main'].get('default_due', '24'))
```
hobarrera: Set the default in the specfile, not here.

tests/test_basic.py
pfertyk: Yeah, sorry for that one :) I was sure that `hours` doesn't work, that was silly :)

todoman/cli.py
pfertyk: Can I use `int` in specfile?
```
default_due = int(default=24)
```

hobarrera: Yes, you can. That should work fine.

tests/test_basic.py
hobarrera: It happens. ^^

todoman/cli.py
pfertyk: It doesn't work for not specified value:
```
  File "/home/pawel/python/todoman/todoman/cli.py", line 152, in new
    default_due = int(ctx.obj['config']['main'].get('default_due'))
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
```

untitaker: I think `integer` should be used, not `int`

hobarrera: Looking at that trace, I'm not sure _what_ isn't working. If you set it to `int(default=24)`, it should never be `None`, but you also don't need to re-cast here.

hobarrera: Yes, @untitaker is right. Ignore my comment, I posted it before I saw his response.

http://configobj.readthedocs.io/en/latest/validate.html#the-standard-functions

pfertyk: @untitaker `integer` seems to work, will push the changes in a moment.

pfertyk: Fixed.

pfertyk:
Is it OK now?

todoman/cli.py
```diff
@@ -149,7 +149,7 @@ def new(ctx, summary, list, todo_properties, interactive):
     if not todo.summary:
         raise click.UsageError('No SUMMARY specified')
 
-    default_due = int(ctx.obj['config']['main'].get('default_due'))
+    default_due = ctx.obj['config']['main'].get('default_due')
```
hobarrera: `default_due = ctx.obj['config']['main']['default_due']` better stays in line with other usages.

tests/test_basic.py
```diff
@@ -192,6 +192,32 @@ def test_default_list(tmpdir, runner, create):
...
 
+@pytest.mark.parametrize(
+    'default_due, expected_due_hours', [(None, 24), (1, 1), (0, None)],
+    ids=['not specified', 'greater than 0', '0']
+)
+def test_default_due(
+    tmpdir, runner, create, default_due, expected_due_hours
+):
+    """
+    Test setting the due date using the default_due config parameter
```
hobarrera: One-line docs should have quotes in the same line:

 `"""Some doc"""`

todoman/cli.py
```diff
@@ -148,6 +149,10 @@ def new(ctx, summary, list, todo_properties, interactive):
     if not todo.summary:
         raise click.UsageError('No SUMMARY specified')
 
+    default_due = ctx.obj['config']['main'].get('default_due')
+    if default_due:
+        todo.due = todo.created_at + timedelta(minutes=default_due * 60)
```
hobarrera: `hours=default_due`?

hobarrera:
LGTM

hobarrera:
Thanks!

pfertyk:
No problem :) When will the new version be available via pip?

hobarrera:
> No problem :) When will the new version be available via pip?

A new one is up (TWO actually, since the first was mispackaged).

PR #83, opened 2017-02-06
Use the system date format by default

hobarrera:
The current date format was ISO8601 (which is great!), but it's honestly a matter of preference. Use the system's default date format if none is specified, since it seems saner to respect the user's system settings that impose out own preference.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,11 @@ Changelog
 This file contains a brief summary of new features and dependency changes or
 releases, in reverse chronological order.
 
+v2.2.0
+------
+
+* Use the system's date format as a default.
+
 v2.1.0
 ------
 
--- a/todoman/cli.py
+++ b/todoman/cli.py
@@ -1,5 +1,6 @@
 import functools
 import glob
+import locale
 from datetime import timedelta
 from os.path import expanduser, isdir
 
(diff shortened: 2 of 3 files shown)


untitaker:
I don't know what that is on my machine. Having system-dependent default makes it harder to write tutorials that everybody can copy-paste from.

hobarrera:
> I don't know what that is on my machine.

Probably related to your language settings? Or the default? I don't honestly see this as a downside.

* If you've never changed it, it'll be whatever distros use as a default.
* If it's something else, it's probably what you see elsewhere.
* Users who know what this value is will be glad to see it being respected.

> Having system-dependent default makes it harder to write tutorials that everybody can copy-paste from.

We don't seem to use these formats for parsing, so it should not break any commands/input*. Users which see different outputs should not really be confused (I _hope_).

* I actually noticed the UI is stricter with datetime formatting than the CLI, that really needs some love.

---

As an alternative, I single config flag like `user_system_locale` might cover it.

untitaker:
We do use those formats for parsing when creating new tasks. The UI is using the same code as the CLI as far as I recall.

hobarrera:
Any veredict on close/merge here?

untitaker:
I don't think I care, I override it anyway :)

PR #84, opened 2017-02-06
WIP: Auto-push to packagecloud.io

hobarrera:
Push Ubuntu/Debian packages to packagecloud.io.

---

Note: This is WIP, since I need to actually test these in a VM.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,9 +9,13 @@ python:
   - 3.6
 
 env:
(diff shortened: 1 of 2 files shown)


hobarrera:
I'm closing this, since packages generated by fpm inside a virtualenv are useless. We probably need another tool (since having native debian machines with each version is unlikely).

PR #86, opened 2017-02-09
Add shortcut for saving in interactive mode

untitaker:


todoman/ui.py
```diff
@@ -91,7 +91,8 @@ def __init__(self, todo, databases, formatter):
         self._help_text = urwid.Text(
             '\n\nGlobal:\n'
             ' F1: Toggle help\n'
-            ' Ctrl-C: Cancel\n\n'
+            ' Ctrl-C: Cancel\n'
+            ' Ctrl-S: Save (only works if not a shell shortcut already)\n\n'
```
hobarrera: Is this commonly a used shell shortcut?

untitaker: Bash overrides it for some obscure feature IIRC.

ctrl-x ctrl-s is used in emacs to save the open file, but I found it unnecessary to emulate that.

hobarrera: Since `bash` overrides <kbd>ctrl</kbd> + <kbd>s</kbd> by default, I'd pick some other binding (`zsh` does this too, it looks pretty standard).

untitaker: Yeah, the thing is, which shortcut makes sense? Ideally it should work within textfields.

hobarrera: All the good ones seem taken I don't think we'll find anything that makes sense. Some options: 

* <kbd>ctrl</kbd> + <kbd>q</kbd>: "save and Quit".
* <kbd>ctrl</kbd> + <kbd>t</kbd>: save (not real relation between action and key).
* <kbd>ctrl</kbd> + <kbd>V</kbd>: "saVe"

untitaker: I'd just bite the bullet and make it `ctrl s`. You can disable bash's usage of that combo and that's what I do anyway because I need that key for mutt.

hobarrera: The thing is, it will be broken by default for almost everyone (`zsh` also binds that by default). I kinda dislike the idea of having a hotkey that won't work by default for almost everyone (we can make them configurable if you *really* want to use <kbd>ctrl</kbd> + <kbd>s</kbd>).

untitaker: At some point we probably should have configurable key bindings, I agree. However, consider that emacs uses `ctrl s` as well, despite this.

hobarrera:
Please update the changelog, and then feel free to merge this (you might need to rebase).

CHANGELOG.rst
```diff
@@ -14,6 +14,7 @@ v2.1.0
 * Attempt to create the cache file's directory if it does not exist.
 * Fix crash when running ``--porcelain show``.
 * Show ``id`` for todos everywhere (eg: including new, etc).
+* Add the ``ctrl-s`` shortcut for saving in the interactive editor.
```
hobarrera: These entries should be in reverse chronological order. :(

untitaker:
Sorry, fixed in master

PR #87, opened 2017-02-09
Show ID everywhere

untitaker:
Given that the ID is now persistent, it makes sense to always print it.

If I create a task and want to edit it immediately, I first have to list
all tasks to get the ID.

### DIFF
--- a/todoman/model.py
+++ b/todoman/model.py
@@ -362,17 +362,18 @@ class Cache:
     def __init__(self, path):
         self.cache_path = str(path)
 
-        self.conn = sqlite3.connect(self.cache_path)
-        self.conn.row_factory = sqlite3.Row
-        self.conn.execute("PRAGMA foreign_keys = ON")
+        self._conn = sqlite3.connect(self.cache_path)
+        self._conn.row_factory = sqlite3.Row
+        self._conn.execute("PRAGMA foreign_keys = ON")
 
+        self.cursor = self._conn.cursor()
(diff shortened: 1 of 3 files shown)


hobarrera:
Thanks, I was thinking the same thing today.

untitaker:
I'll fix tests

hobarrera:
Oh, yeah, now I remember why I didn't implement it; when reading from a file, the `id` is not set. We'd need to implement this somewhere (IIRC, `new` is especially painful).

todoman/model.py
```diff
@@ -781,6 +787,16 @@ def flush(self):
...
 
+    def save(self, todo, list_):
+        todo.save(list_)
+        path = os.path.join(list_.path, todo.filename)
+        self.cache.expire_file(path)
+        mtime = _getmtime(path)
+        self.cache.add_file(list_.name, path, mtime)
+        id = self.cache.add_todo(todo, path)
+        self.cache.save_to_disk()
+        todo.id = id
```
untitaker: This just patches the `id` onto `FileTodo`. Is this OK?

hobarrera: Yeah, I can't really think of any other way to do this.

untitaker:
`lastrowid` (which we return in `add_todo`) is only available with a cursor

hobarrera:
Oh, okay. LGTM, feel free to merge this.

PR #88, opened 2017-02-11
Stop using setuptools' [slow] entry points

hobarrera:
setuptools creates really slow scripts (especially for the system python). Our own small script shows a considerable startup speed increase.

On my system, this drops speed from ~850ms to 200ms. I'd love feedback on this (both implementation and results).


untitaker:
Strongly recommending against this hack. Better to place effort into optimizing setuptools, see https://github.com/pypa/setuptools/pull/901#issuecomment-269880907

hobarrera:
While I agree that fixing setuptools is a nice long-term goal, I feel that it moves rather slowly, and we can use this in the meantime, and eventually move back to setuptools when it gets fixed.

untitaker:
Oh well. This just looks very hacky to me in any case, and I wonder if we won't break somebody's usecase (or distro packaging routine) just by doing this.

Have you considered using the `scripts` of `setuptools.setup`?

hobarrera:
> Have you considered using the scripts parameter of setuptools.setup?

Yes. I was my first approach, and had the same issue.

TBH, I'm not entirely convinced by this either, it *is* quite hacky, but the speed gains are gigantic (I'd also like to confirm it's not just me). I honestly hate having to choose between performance and correctness.

hobarrera:
Using `scripts` generates this:

```
$ cat $(which todo)
#!/usr/bin/python
# EASY-INSTALL-SCRIPT: 'todoman==2.0.2.post4+ng5ae8446.d20170211','todo'
__requires__ = 'todoman==2.0.2.post4+ng5ae8446.d20170211'
__import__('pkg_resources').run_script('todoman==2.0.2.post4+ng5ae8446.d20170211', 'todo')
```

untitaker:
How does your `setup.py` look like?

hobarrera:
The `setup.py` used in the above test:

```
#!/usr/bin/env python3

from setuptools import setup

setup(
    name='todoman',
    description='A simple CalDav-based todo manager.',
    author='Hugo Osvaldo Barrera',
    author_email='hugo@barrera.io',
    url='https://github.com/pimutils/todoman',
    license='MIT',
    packages=['todoman'],
    include_package_data=True,
    # data_files=[
    #     (os.path.dirname(sys.executable), ['bin/todo']),
    # ],
    scripts=[
        'bin/todo',
    ],
    install_requires=[
        open('requirements.txt').readlines()
    ],
    long_description=open('README.rst').read(),
    use_scm_version={
        'version_scheme': 'post-release',
        'write_to': 'todoman/version.py',
    },
    setup_requires=['setuptools_scm != 1.12.0', 'pytest-runner'],
    tests_require=['pytest'],
    classifiers=[
        'Development Status :: 4 - Beta',
        'Environment :: Console',
        'Environment :: Console :: Curses',
        'License :: OSI Approved :: MIT License',
        'Operating System :: POSIX',
        'Programming Language :: Python :: 3',
        'Programming Language :: Python :: 3.3',
        'Programming Language :: Python :: 3.4',
        'Programming Language :: Python :: 3.5',
        'Programming Language :: Python :: 3.6',
        'Topic :: Office/Business :: Scheduling',
        'Topic :: Utilities',
    ]
)
```

hobarrera:
FWIW, the script is actually placed in `/usr/lib/python3.6/site-packages/todoman-2.0.2.post4+ng5ae8446.d20170211-py3.6.egg/EGG-INFO/scripts`.

untitaker:
I suspect this will break when using the `--prefix` option of `pip install`. Perhaps that path can be determined from within setup.py.

hobarrera:
Yes, that's what I meant (and what I thought you were referring to before).

hobarrera:
After some thinking about this a bit more, I've decided there's a simpler road for this:

* Include a `bin/todo` script[1].
* Document that users should install via `pip` (eg: not `setup.py install`) (which will use the wheel and we've been told doesn't have the same issues).
* Document that packagers should actually overwrite setuptools' entry point with `bin/todo`.

[1]:
```python
from todoman import cli

if __name__ == '__main__':
    cli.cli()
```

hobarrera:
How do you feel about this adaptation, @untitaker ?

Also, do you think we can do something similar for khal?

bin/todo
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env python
```
untitaker: This is `python2` on most systems.

docs/source/install.rst
```diff
@@ -35,12 +35,32 @@ Requirements
...
+
+Notes for Packagers
+-------------------
+
+All of todoman's dependencies are listed in the requirements.txt_ file. New
+dependencies will be clearly announced in the ``CHANGELOG.rst`` file for each
+release. Patch releases (eg: those where the third digit of the version is
+incremented) **will not** introduce new dependencies.
+
+If your packages are generated by running ``setup.py install`` or some similar
```
untitaker: Is that true? I thought this only affects `python setup.py develop`.

hobarrera: Yup, also affects `setup.py install`. The AUR package currently uses this and takes ~1s extra comparted to this script.

bin/todo
hobarrera: Since this is mostly for developers, it's up to them to patch this for any other python names. Keep in mind that setup.py also uses this hashbang.

We might want to document this detail though.

PR #93, opened 2017-02-13
Fix due datetime tz-awareness comparison

### LINKED ISSUE #92: sqlite due comparisons don't consider timezones
Since dates are stored as strings in sqlite, timezones aren't taking properly accounted for when comparing datetimes. (this was revealed by df24af89115b02e588ac0112aa9160719621f362)

hobarrera:
Fixes #92

### DIFF
--- a/tests/test_filtering.py
+++ b/tests/test_filtering.py
@@ -1,7 +1,10 @@
 from datetime import datetime, timedelta
 
+import pytz
+from dateutil.tz import tzlocal
+
 from todoman.cli import cli
-from todoman.model import Database
+from todoman.model import Database, FileTodo
 
 
 def test_all(tmpdir, runner, create):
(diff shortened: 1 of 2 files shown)


hobarrera:
*Sigh*. I guess we can't drop python < 3.6 **yet**.

tests/test_filtering.py
```diff
@@ -162,22 +165,25 @@ def test_filtering_lists(tmpdir, runner, create):
...
+
     todos = list(db.todos(due=24))
 
-    assert len(todos) == 2
+    assert len(todos) == 4
     assert todos[0].summary == "23"
-    assert todos[1].summary == "1"
+    assert todos[1].summary == "23"
+    assert todos[2].summary == "1"
+    assert todos[3].summary == "1"
```
untitaker: This part is a bit awkward. Which contract guarantees this particular order of the `db.todos()` iterator?

todoman/model.py
```diff
@@ -375,7 +377,35 @@ def __init__(self, path):
     def save_to_disk(self):
         self._conn.commit()
 
+    def _check_version(self):
```
untitaker: Could be called `is_latest_version`

tests/test_filtering.py
hobarrera: `db.todos()` receives a bunch of parameters to specify an explicit order, and otherwise guarantees the default order. It's actually the place where our app-wide order is defined. ;)

todoman/model.py
hobarrera: Yes, definitely.

PR #99, opened 2017-02-15
Strip the human_time flag

hobarrera:
Always parse using human times. Scripts and integrations can use --porcelain.

PR is a lot shorter with [`?w=1`](https://github.com/pimutils/todoman/pull/99/files?w=1)

@untitaker Objections?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,8 @@ releases, in reverse chronological order.
 v2.1.0
 ------
 
+* The ``--no-human-time`` flag is gone. Integrations/scripts might want to look
+  at ``--porcelain`` as an alternative.
 * Fix crash when running ``todo new``.
 * Fixes some issues when filtering todos from different timezones.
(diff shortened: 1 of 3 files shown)


untitaker:
I don't like the idea that I cannot disable human time parsing when using
todoman in scripts

hobarrera:
Why not `porcelain` for scripts?

hobarrera:
Also, would having `human-time` disable break them? I'd expect not.

hobarrera:
Test failure is totally unrelated, and a timing issue in how the test was written:
https://travis-ci.org/pimutils/todoman/jobs/202052596

untitaker:
Porcelain doesn't disable human times for todo new, right?

hobarrera:
It sort of does: it replaces the default formatter, so all formatting and parsing uses ISO-8601.

untitaker:
Ah, nevermind then, lgtm

PR #108, opened 2017-02-19
Use tomorrow and today while formatting date

### LINKED ISSUE #82: Use "tomorrow" and "today" when formatting date
When a todo expires today, or tomorrow, show "today" or "tomorrow" respectively instead of the actual date.

rimshaakhan:
fixes #82


todoman/ui.py
```diff
@@ -251,8 +251,23 @@ def detailed(self, todo):
...
 
     def format_date(self, date):
+        """
+        Returns date in the following format:
+        * if date == today or tomorrow: "Today" or "Tomorrow"
+        * else: return a string representing that date
+        """
         if date:
-            rv = date.strftime(self.date_format)
+            date_tomorrow = datetime.today().date() + timedelta(days = 1)
```
untitaker: Instead of calling `datetime.today()` here, please use `self.now`. This is because we might want to replace `self.now` with a fixed value in the testsuite.

todoman/ui.py
```diff
@@ -251,8 +251,23 @@ def detailed(self, todo):
...
+            # Get a dictionary which stores the dates for which we don't have to merely
+            # return a date string and map it to the special string we need to return
+            dates_for_which_not_to_return_simple_date = {
```
untitaker: This is quite a long variable name, I suggest something like `special_dates`.

todoman/ui.py
```diff
@@ -251,8 +251,23 @@ def detailed(self, todo):
...
+                datetime.today().date() : "Today",
+                date_tomorrow : "Tomorrow" ,
+            }
+            if date in dates_for_which_not_to_return_simple_date.keys():
```
untitaker: `.keys()` is not necessary, just `if date in dates_for_which_not_to_return_simple_date`.

rimshaakhan:
Thanks @untitaker for your kind help!

Looking forward to correct my mistakes

rimshaakhan:
Hello @untitaker, do I need to write a test case for testing this feature?

todoman/ui.py
```diff
@@ -251,8 +251,24 @@ def detailed(self, todo):
...
-            rv = date.strftime(self.date_format)
+            date_tomorrow = self.now.date() + timedelta(days=1)
+            # Get a dictionary which stores the dates for which
+            # we don't have to merely return a date string and
+            # map it to the special string we need to return
+            special_dates = {
+                self.now.date(): "Today",
+                date_tomorrow: "Tomorrow",
+            }
+            if date in special_dates():
```
untitaker: You need to write `special_dates` instead of `special_dates()`.

rimshaakhan: Oops! Sorry

untitaker:
>do I need to write a test case for testing this feature?

That would be nice. You have two options:

1. Write a test that simulates running `todo`, it would belong into https://github.com/pimutils/todoman/blob/master/tests/test_basic.py

  For that option I imagine it would be enough to run `todo new --due tomorrow hello` and check if `todo list` outputs `Tomorrow` at some point. Also a similar testcase for `Today`.

2. Unit tests for the formatter. We currently have none, so this would require creating a new file `tests/test_formatter.py`. In the test you could create an instance of `TodoFormatter` (with a datetime format string of your choice), and test `format_date` with some input

todoman/ui.py
untitaker: No need to apologize!

hobarrera:
Hi. Thanks for picking this up with be being a bit away, @untitaker.

Another detail is that, if you run `todo`, the outputs are aligned, so `today` and `yesterday` are probably need to have an *exact* width. This is why no-date returns `empty_date`, which is a string with spaces to fill up the column.

You can probably do something like `"today".rjust(WIDTH)`. I'd do this is the constructor (`__init__`) so we only compute this once (this function is called once per todo).

hobarrera:
> It resolves Issue #82

Hint: if the message says something like "closes #82" or "fixes #82", merging the PR will automatically close the issue. ;)

rimshaakhan:
The line which is making flake8 violations is [here](https://github.com/rimshaakhan/todoman/blob/master/todoman/ui.py#L270):
```
return special_dates[date] if date in special_dates else date.strftime(self.date_format)
```
Isn't it the optimal way of doing it or should I switch back to the if-else tree?

untitaker:
@rimshaakhan The line is too long. I personally prefer the if-else variant.

todoman/ui.py
```diff
@@ -251,9 +251,23 @@ def detailed(self, todo):
...
+        * if no date is supplied, it returns empty_date
+        """
         if date:
-            rv = date.strftime(self.date_format)
-            return rv
+            date_tomorrow = self.now.date() + timedelta(days=1)
+            # Get a dictionary which stores the dates for which
+            # we don't have to merely return a date string and
+            # map it to the special string we need to return
+            WIDTH = len(date.strftime(self.date_format))
```
hobarrera: Actually, `len(empty_date)` was fine. They should both have the same with, and we avoid formatting this date when it's [possibly] unnecessary.

I'd also prefer move this over to `__init__` , so we only calculate  this width once, instead every time we format a date (pretty much like we only calculate `empty_date` once.

Let me know if you're not very familiar with contructors/`__init__` or if this is otherwise unclear.

Also, `date_width` is actually a better variable name, but that one's my fault because I used `WIDTH` on my example, which really was a bad idea on my part, sorry.

todoman/ui.py
```diff
@@ -251,9 +251,23 @@ def detailed(self, todo):
...
+            special_dates = {
+                self.now.date(): "Today".rjust(WIDTH, " "),
```
hobarrera: Maybe the same as above; define the strings in the constructor.

rimshaakhan:
No, I am not feeling demotivated, but on the contrary it  gets even more exciting and adventurous.
It's feeling fun and on a serious note: I need to learn all these stuff, when else could have been a better time? - I am really grateful for having such technically insightful and creative critique on my work!
Thank You

todoman/ui.py
rimshaakhan: I guess you would want something like this:
```
def __init__(self, date_format):
    self.date_format = date_format
    self._localtimezone = tzlocal()
    self.now = datetime.now().replace(tzinfo=self._localtimezone)
    # Get a dictionary which stores the dates for which
    # we don't have to merely return a date string and
    # map it to the special string we need to return
    self.special_dates = {
        self.now.date(): "Today".rjust(self.date_width, " "),
        date_tomorrow: "Tomorrow".rjust(self.date_width, " "),
    }
    # An empty date which should be used in case no date is present
    self.empty_date = " " * len(self.format_date(self.now))
    self.date_width = len(self.empty_date)
    self._parsedatetime_calendar = parsedatetime.Calendar()
```
But then wouldn't  ```self.empty_date``` call ```self.format_date```, where ```self.date_width``` wouldn't be present and there would be an exception.
Should we find a different way for calculating ```self.empty_date```  or should we leave date_width in function itself?

hobarrera: Ugh, you're completely right, I missed that! I think we need to rewrite how we calculate `empty_date` and avoid that trap:

```python
empty_date = self.now.format(self.date_format)
```

I don't think we need the comment explaining what that dictionary is; it seems quite clear. Maybe just a short one-liner, at most.

rimshaakhan: Also, since ```self.empty_date``` calls ```self.format_date(self.now))```, it would not be an empty_date, rather something like "        Today" (ideally - when we can get the code to run) and I think it would hinder the utility of having an empty_date and therefore would render it useless! :(

hobarrera: Oops (again!). Not by best day. I meant this:
```python
self.date_width = len(self.now.format(date_format))
self.empty_date = " " * self.date_width

```

hobarrera:
FWIW, these final items are listed in the [patch review checklist](https://todoman.readthedocs.io/en/stable/contributing.html#patch-review-checklist).

todoman/ui.py
```diff
@@ -204,8 +204,16 @@ def __init__(self, date_format):
         self.date_format = date_format
         self._localtimezone = tzlocal()
         self.now = datetime.now().replace(tzinfo=self._localtimezone)
-        self.empty_date = " " * len(self.format_date(self.now))
-
+        self.date_tomorrow = self.now.date() + timedelta(days=1)
```
hobarrera: Just call this `self.tomorrow`, for consistency.

rimshaakhan:
What should be added to changelog?

Perhpas, just a line saying: ```* Proper date formatting enabled ``` would be enough!

untitaker:
I would rather describe the concrete changes to `todoman`'s behavior. The PR title is pretty close to that, though the word formatting may be too technical for the changelog, which is supposed to be a summary targeted at users.

todoman/ui.py
```diff
@@ -251,8 +259,18 @@ def detailed(self, todo):
         return rv
 
     def format_date(self, date):
+        """
+        :param datetime.date date: a datetime object
```
untitaker: In fact `format_date` recieves a datetime object (i.e. `datetime.datetime` instead of `datetime.date`). It is probably useful to add an `assert isinstance(date, datetime)` in `format_date`.

todoman/ui.py
```diff
@@ -251,8 +259,18 @@ def detailed(self, todo):
...
+        Returns date in the following format:
+        * if date == today or tomorrow: "Today" or "Tomorrow"
+        * else: return a string representing that date
+        * if no date is supplied, it returns empty_date
+        """
         if date:
-            rv = date.strftime(self.date_format)
+            if date in self.special_dates:
```
untitaker: Since `date` is a datetime (I know, misleading variable name), you need to use `date.date()` here, otherwise `special_dates` is never used.

tests/test_formatter.py
```diff
@@ -0,0 +1,20 @@
...
+from todoman import ui
+
+DATE_FORMAT = "%d-%m-%y"
+
+def test_format_date():
+    """
+    Tests the format_date function in todoman.ui.TodoFormatter
+    """
+    formatter = ui.TodoFormatter(DATE_FORMAT)
+    today = datetime.now().date()
```
untitaker: Please use `datetime` here

todoman/ui.py
hobarrera: Yes, sorry, this is something we've dragged for a while. #109 will fix this in future, but for now, just update this to `date.date()`.

untitaker:
@rimshaakhan if you run `tox -epy` locally, you will see that your current testcase fails. Also there are a few styling problems with your latest commits (see `tox -eflake8`)

rimshaakhan:
> @rimshaakhan if you run tox -epy locally, you will see that your current testcase fails. Also there are a few styling problems with your latest commits (see tox -eflake8)

I ran the tox test, but there was an error: Here's the log file: [epy_test.txt](https://github.com/pimutils/todoman/files/789845/epy_test.txt)

Can you please take a look at it! (I have already removed the pyc files)

hobarrera:
This is a bit odd, and I'm not 100% certain what went wrong. Can you delete `.tox`, and retry running `tox -e py35`? Thanks.

untitaker:
#113 works around this

tests/test_formatter.py
```diff
@@ -1,20 +1,21 @@
...
-    any_day = today + datetime.timedelta(days = random.randrange(2))
-
-    str_format = lambda x: x.rjust((formatter.date_width), " ")
+    any_day = today + timedelta(days = randrange(2, 8))
 
-    assert formatter.format_date(today) == str_format("Today")
-    assert formatter.format_date(tomorrow) == str_format("Tomorrow")
-    assert formatter.format_date(any_day) == any_day.strftime(DATE_FORMAT)
\ No newline at end of file
+    assert formatter.format_date(today) == "Today".rjust(date_width)
```
hobarrera: It's better use `'   Today'` here.  That way, if somebody breaks how `date_width` is calculated in future, this test fill fail and let us know about it immediately.

rimshaakhan:
> This is a bit odd, and I'm not 100% certain what went wrong. Can you delete .tox, and retry running tox -e py35? Thanks.

Nope! Nothing happened, same error

hobarrera:
Yeah, @untitaker narrowed it down and fixed it in #113.

Basically, pytest is using tests from the virutalenv (pytest just looks for anything that looks like a test inside the current directory, recursively), instead of the real ones, and that confuses it.

AUTHORS.rst
```diff
@@ -15,3 +15,4 @@ We manually track this since commit author *does not* equal code author.
 * Markus Unterwaditzer <markus@unterwaditzer.net>
 * Paweł Fertyk <pfertyk@openmailbox.org>
 * Thomas Glanzmann <thomas@glanzmann.de>
+* Rimsha Khan <rimshakhan.rk03@gmail.com>
```
hobarrera: This list is kept in alphabetical order (I guess I should clarify that).

rimshaakhan:
I am quite unfamiliar with using pytest, and I cannot understand why the current CI build failed.

Is this something I broke?

hobarrera:
Uhm, sort of (but this was expected!).

We had some tests that created a todo, and checked that the output was the right one (like [this one](https://github.com/pimutils/todoman/blob/f3e29d689704a8e61e31e5fd3c6e80e40e33843d/tests/test_basic.py#L334-L353)).

Since the format of the date changed, these tests broke (all they do is reflect that the output is not what we expected at the time of their writing).

The simple fix, is to change the the `[-1, -1]` on top to `[-72, 72]`, so your change does not affect it.

todoman/ui.py
```diff
@@ -251,8 +259,19 @@ def detailed(self, todo):
...
 
     def format_date(self, date):
+        """
+        :param datetime.datetime date: a datetime object
+        Returns date in the following format:
+        * if date == today or tomorrow: "Today" or "Tomorrow"
+        * else: return a string representing that date
+        * if no date is supplied, it returns empty_date
+        """
+        assert isinstance(date, datetime)
```
hobarrera: This needs to be moved one line down:
```python
if date:
    assert(...
```

Otherwise it fails when date is `None` (in which case we'd return `empty_date`).

rimshaakhan: Oh! So sorry, I completely forgot about it!

Should I also add a test case where it would check for the case when format_date receives ```None```

hobarrera: Yes please. The test should expect that the result is `'        '` (note: I didn't count the spaces, I might be off)

rimshaakhan:
>Yeah, @untitaker narrowed it down and fixed it in #113.
Basically, pytest is using tests from the virutalenv (pytest just looks for anything that looks like a test inside the current directory, recursively), instead of the real ones, and that confuses it.

I have pulled in changes, but tox stil shows the same [error log](https://github.com/pimutils/todoman/files/791235/tox_output.txt) :(

untitaker:
Sorry, I should've tested #113 properly. #114 incoming 😕

rimshaakhan:
>Sorry, I should've tested #113 properly. #114 incoming :confused:

That worked! :tada:

rimshaakhan:
The color checking te:
```python
@pytest.mark.parametrize('hours', [72, -72])
def test_color_due_dates(tmpdir, runner, create, hours):
    due = datetime.datetime.now() + datetime.timedelta(hours=hours)
    create(
        'test.ics',
        'SUMMARY:aaa\n'
        'STATUS:IN-PROGRESS\n'
        'DUE;VALUE=DATE-TIME;TZID=ART:{}\n'
        .format(due.strftime('%Y%m%dT%H%M%S'))
    )

    result = runner.invoke(cli, ['--color', 'always'])
    assert not result.exception
    due_str = due.strftime('%Y-%m-%d')
    if hours == 1:
        assert result.output == \
            '  1 [ ]   {} aaa @default\x1b[0m\n'.format(due_str)
    else:
        assert result.output == \
            '  1 [ ]   \x1b[31m{}\x1b[0m aaa @default\x1b[0m\n'.format(due_str)
```
is failing with the error message:

```
            '  1 [ ]   {} aaa @default\x1b[0m\n'.format(due_str)
    else:
>       assert result.output == \
            '  1 [ ]   \x1b[31m{}\x1b[0m aaa @default\x1b[0m\n'.format(due_str)
E       assert '  1 [ ]   20...ault\x1b[0m\n' == '  1 [ ]   \x1...ault\x1b[0m\n'
E         -   1 [ ]   2017-02-25 aaa @default
E         +   1 [ ]   2017-02-25 aaa @default
E         ?           +++++          ++++

tests/test_basic.py:353: AssertionError

```

Can someone please share any insight @hobarrera @untitaker ?
Thanks!

untitaker:
Yes, in that test there is a line called `if hours == 1`, it also needs to be changed to `if hours == 72`

CHANGELOG.rst
```diff
@@ -23,6 +23,7 @@ v2.1.0
 * Fix crash when running ``--porcelain show``.
 * Show ``id`` for todos everywhere (eg: including new, etc).
 * Add the ``ctrl-s`` shortcut for saving in the interactive editor.
+* Show "Today" or "Tomorrow" when due date is today or tomorrow
```
hobarrera: v2.1.0 was released a couple of days ago, so you need to move this up. ^_^

Note that I try and keep things in reverse-chronological order too, so it should be at the very top.

todoman/ui.py
```diff
@@ -251,8 +259,19 @@ def detailed(self, todo):
         return rv
 
     def format_date(self, date):
+        """
+        :param datetime.datetime date: a datetime object
```
hobarrera: `:param:` need to be at the bottom of the docstring, not at the top.

tests/test_formatter.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+def test_format_date():
+    """
+    Tests the format_date function in todoman.ui.TodoFormatter
+    """
+    formatter = ui.TodoFormatter(DATE_FORMAT)
+    today = datetime.now()
+    tomorrow = today + timedelta(days = 1)
+    any_day = today + timedelta(days = randrange(2, 8))
```
untitaker: So close! A styleguide violation here, no spaces around the `=`. Same for the previous line.

rimshaakhan: Hi @untitaker, I am using Sublime Text 3
It was so close! :)

untitaker:
I think [this](https://github.com/pimutils/todoman/pull/108#pullrequestreview-23080556) review comment is still pending, but then that'd be really it :)

rimshaakhan:
Also, should I consider filing a issue because I am unable to filter out results when I run ```flake8``` , it reports a issues in the env files.

Please have a look at the [log file](https://github.com/pimutils/todoman/files/791665/flake_output.txt)

tests/test_formatter.py
hobarrera: Back when I used sublime I used a plugin called `Anaconda` that did some python style checks and stuff (don't recall what else). Might be worth looking into.

hobarrera:
> Also, should I consider filing a issue because I am unable to filter out results when I run flake8 , it reports a issues in the env files.

Run `tox -e flake8` to only check `todoman` files.

rimshaakhan:
> Run tox -e flake8 to only check todoman files.

Yeah! that was the command I ran, but didn't quite well worked for me!
Perhaps, it is because I have named the virtualenv as todoman_virtualenv (I may sound like a complete idiot )

hobarrera:
No, it's not really anything you've done. Personally, I never came across that because I use `mkvirtualenv` (which creates all the virtualenvs in a separate location). I'm not sure how other projects deal with that though.

hobarrera:
Thanks for you contribution! 🎉 
Hope this has been useful and a worthwhile learning experience!

untitaker:
In general I would recommend creating the virtualenv outside of the git repository, otherwise you may accidentally commit it into git, or misconfigured tools may crawl it like pytest or flake8.

untitaker:
In general I would recommend creating the virtualenv outside of the git repository, otherwise you may accidentally commit it into git, or misconfigured tools may crawl it like pytest or flake8.

rimshaakhan:
@hobarrera You just beat me up while I was writing this Thank You  :)
@untitaker @hobarrera Thank You sir for being so wonderfully patient and for giving me critical reviews.

I am looking forward to apply for RGSoC and am very excited to learn: how to make fascinating code which looks so amazingly beautiful!

Its a wonderful feeling of seeing myself in the author list
Thank You

untitaker:
@rimshaakhan "beat me up" doesn't mean what you think it does! :D It has been a pleasure working with you so far!

rimshaakhan:
>In general I would recommend creating the virtualenv outside of the git repository, otherwise you may accidentally commit it into git, or misconfigured tools may crawl it like pytest or flake8.

@untitaker Should I re-configure for the same!
I did thought of it earlier, but couldn't think of a possible solution! (There's mkvirtualenv now :) as @hobarrera  pointed out)

PR #113, opened 2017-02-21
Narrow down pytest search path

untitaker:
See https://github.com/pimutils/todoman/pull/108#issuecomment-281426993


hobarrera:
Aaah, that makes a bunch of sense. I use `mkvirtualenv`, so never came across the underlying issue here. Be sure to add a quick not into the commit description (possible when squash-merging).

PR #117, opened 2017-02-22
Support due times

### LINKED ISSUE #109: Support times for due dates
- Todoman should introduce a `time_format` configuration parameter that, if non-empty, is *additionally* used to format due datetimes. It should be omitted if `DUE` does not have a time component.
- Todoman should prevent people from using `%H`, %M` and so on in `date_format` (I'm currently doing that)
- Likewise it should prevent people from using `%Y` etc in `time_format`

untitaker:
Fix #109

dt_separator is necessary because of parsing due dates

### DIFF
--- a/tests/test_formatter.py
+++ b/tests/test_formatter.py
@@ -4,18 +4,23 @@ from random import randrange
 from todoman import ui
 
 DATE_FORMAT = "%d-%m-%y"
+TIME_FORMAT = "%H:%M"
 
 
-def test_format_date():
-    """
-    Tests the format_date function in todoman.ui.TodoFormatter
-    """
-    formatter = ui.TodoFormatter(DATE_FORMAT)
+def test_human_dates():
+    formatter = ui.TodoFormatter(DATE_FORMAT, TIME_FORMAT, ' ')
     today = datetime.now()
     tomorrow = today + timedelta(days=1)
+    tomorrow_12am = datetime(tomorrow.year, tomorrow.month, tomorrow.day,
+                             12, 0)
+    today_12am = datetime(today.year, today.month, today.day,
+                          12, 0)
     any_day = today + timedelta(days=randrange(2, 8))
 
-    assert formatter.format_date("") == "        "
-    assert formatter.format_date(today) == "   Today"
-    assert formatter.format_date(tomorrow) == "Tomorrow"
-    assert formatter.format_date(any_day) == any_day.strftime(DATE_FORMAT)
+    assert formatter.format_datetime("") == "              "
+    assert formatter._format_date(today.date()) == "   Today"
+    assert formatter._format_date(tomorrow.date()) == "Tomorrow"
+    assert formatter.format_datetime(any_day) == \
+        any_day.strftime(DATE_FORMAT + ' ' + TIME_FORMAT)
+    assert formatter.format_datetime(tomorrow_12am) == "Tomorrow 12:00"
+    assert formatter.parse_datetime('12:00').replace(tzinfo=None) == today_12am
--- a/todoman/configuration.py
+++ b/todoman/configuration.py
@@ -33,6 +33,24 @@ def validate_cache_path(path):
         )
 
 
+def validate_date_format(fmt):
+    if any(x in fmt for x in ('%H', '%M', '%S', '%X')):
+        raise ConfigurationException(
+            'Found time component in `date_format`, please use `time_format` '
+            'for that.'
+        )
+    return fmt
+
+
+def validate_time_format(fmt):
+    if any(x in fmt for x in ('%Y', '%y', '%m', '%d', '%x')):
+        raise ConfigurationException(
+            'Found date component in `time_format`, please use `date_format` '
+            'for that.'
+        )
+    return fmt
+
+
 def validate_color(color):
     color = is_option(color)
     if color == 'always':
@@ -67,6 +85,8 @@ def load_config():
     validator = Validator({
         'expand_path': expand_path,
         'cache_path': validate_cache_path,
+        'date_format': validate_date_format,
+        'time_format': validate_time_format,
     })
 
     config = ConfigObj(path, configspec=specpath, file_error=True)
--- a/todoman/confspec.ini
+++ b/todoman/confspec.ini
@@ -12,7 +12,13 @@ color = option('always', 'auto', 'never', default='auto')
 
 # The date format used both for displaying dates, and parsing input dates. If
 # this option is not specified the system locale's is used.
-date_format = string(default='%x')
+date_format = date_format(default='%x')
+
+# The date format used both for displaying times, and parsing input times.
+time_format = time_format(default='%X')
+
+# The string used to separate date and time when displaying and parsing.
+dt_separator = string(default=' ')
 
 # The default list for adding a todo. If you do not specify this option, you
 # must use the ``--list`` / ``-l`` option every time you add a todo.
--- a/todoman/ui.py
+++ b/todoman/ui.py
@@ -1,5 +1,5 @@
+import datetime
 import json
-from datetime import datetime, timedelta
 from time import mktime
 
 import click
@@ -39,13 +39,13 @@ class TodoEditor:
 
         if todo.due:
             # TODO: use proper date_format
-            due = formatter.format_date(todo.due)
+            due = formatter.format_datetime(todo.due)
         else:
             due = ""
 
         if todo.start:
             # TODO: use proper date_format
-            dtstart = formatter.format_date(todo.start)
+            dtstart = formatter.format_datetime(todo.start)
         else:
             dtstart = ''
 
@@ -143,8 +143,8 @@ class TodoEditor:
         self.todo.summary = self.summary
         self.todo.description = self.description
         self.todo.location = self.location
-        self.todo.due = self.formatter.parse_date(self.due)
-        self.todo.start = self.formatter.parse_date(self.dtstart)
+        self.todo.due = self.formatter.parse_datetime(self.due)
+        self.todo.start = self.formatter.parse_datetime(self.dtstart)
 
         self.todo.is_completed = self._completed.get_state()
 
@@ -200,15 +200,22 @@ class TodoFormatter:
         "{id:3d} [{completed}] {urgent} {due} {summary} {list}{percent}"
     # compact_format = "{completed} {urgent}  {due}  {summary}"
 
-    def __init__(self, date_format):
+    def __init__(self, date_format, time_format, dt_separator):
         self.date_format = date_format
+        self.time_format = time_format
+        self.dt_separator = dt_separator
+        self.datetime_format = date_format + dt_separator + time_format
+
         self._localtimezone = tzlocal()
-        self.now = datetime.now().replace(tzinfo=self._localtimezone)
-        self.tomorrow = self.now.date() + timedelta(days=1)
+        self.now = datetime.datetime.now().replace(tzinfo=self._localtimezone)
+        self.tomorrow = self.now.date() + datetime.timedelta(days=1)
 
         # An empty date which should be used in case no date is present
         self.date_width = len(self.now.strftime(date_format))
         self.empty_date = " " * self.date_width
+
+        self.time_width = len(self.now.strftime(time_format))
+        self.empty_time = " " * self.time_width
         # Map special dates to the special string we need to return
         self.special_dates = {
             self.now.date(): "Today".rjust(self.date_width, " "),
@@ -230,7 +237,7 @@ class TodoFormatter:
             percent = " ({}%)".format(percent)
         urgent = " " if todo.priority in [None, 0] else "!"
 
-        due = self.format_date(todo.due)
+        due = self.format_datetime(todo.due)
         if todo.due and todo.due <= self.now and not todo.is_completed:
             due = click.style(due, fg='red')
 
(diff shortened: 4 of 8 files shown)


todoman/confspec.ini
```diff
@@ -14,6 +14,13 @@ color = option('always', 'auto', 'never', default='auto')
 # this option is not specified the system locale's is used.
 date_format = string(default='%x')
 
+# The date format used both for displaying times, and parsing input times.
+# Default is `%H:%M`.
```
hobarrera: There's no need to clarify this default due to how the doc is rendered (the one from below will be picked up and shown already).

todoman/ui.py
```diff
@@ -258,41 +265,67 @@ def detailed(self, todo):
...
         * if no date is supplied, it returns empty_date
 
-        :param datetime.datetime date: a datetime object
+        :param datetime.date date: a date object
         """
         if date:
-            assert isinstance(date, datetime)
-            if date.date() in self.special_dates:
-                rv = self.special_dates[date.date()]
+            assert type(date) is datetime.date
```
hobarrera: We don't really need this assertion any more due to how we call this function (we've also made it sort of non-public now).

todoman/ui.py
```diff
@@ -258,41 +265,67 @@ def detailed(self, todo):
...
             return rv
         else:
             return self.empty_date
 
-    def parse_date(self, date):
-        if not date:
-            return None
+    def _format_time(self, time):
+        if time:
+            assert isinstance(time, datetime.time)
```
hobarrera: Ditto.

todoman/confspec.ini
```diff
@@ -14,6 +14,13 @@ color = option('always', 'auto', 'never', default='auto')
...
+time_format = string(default='%H:%M')
+
+# The string used to separate date and time when displaying and parsing.
+dt_separator = string(default=' ')
```
hobarrera: As a default, I'd prefer to have:
```python
if time_format:
    return ' '
else:
    return ''
```

This way, we avoid an extra space when there's no `time_format` (I think it's burdensome for users to have to _disable_ the separator when they don't use times.

tests/test_formatter.py
```diff
@@ -4,18 +4,17 @@
...
     tomorrow = today + timedelta(days=1)
     any_day = today + timedelta(days=randrange(2, 8))
 
-    assert formatter.format_date("") == "        "
-    assert formatter.format_date(today) == "   Today"
-    assert formatter.format_date(tomorrow) == "Tomorrow"
-    assert formatter.format_date(any_day) == any_day.strftime(DATE_FORMAT)
+    assert formatter.format_datetime("") == "              "
+    assert formatter._format_date(today.date()) == "   Today"
+    assert formatter._format_date(tomorrow.date()) == "Tomorrow"
```
hobarrera: Can we also have a test for `'   Tomorrow 12:00'`?

todoman/ui.py
```diff
@@ -258,41 +265,67 @@ def detailed(self, todo):
...
-                raise ValueError(
-                    'Time description not recognized: {}' .format(date)
-                )
-            rv = datetime.fromtimestamp(mktime(rv))
+    def format_datetime(self, dt):
+        if not dt:
+            date_part = None
+            time_part = None
+        else:
+            assert isinstance(dt, datetime.datetime)
```
hobarrera: Since we're doing this asserting, can't we do `dt.strtime(self.datetime_format)`? Can a `datetime` *not* have a time component?

To deal with `special_dates`, we can just substitute `'Tomorrow`' with '`Tomorror' + SEPARATOR + TIME_FORMAT` (at construction time).

todoman/confspec.ini
untitaker: If one of `time_format` or `date_format` is an empty string, `dt_separator` is not used. Hence the `filter(bool, ...)` construction.

todoman/ui.py
untitaker: Using `dt.strftime` obliterates the entire special-casing done in #108

todoman/ui.py
untitaker: Done

todoman/ui.py
untitaker: Done

todoman/confspec.ini
untitaker: Done

untitaker:
@hobarrera I think I made all requested changes.

todoman/ui.py
hobarrera: Not the way I was thinking it:

```python
self.special_dates = {
    self.now.date(): "today".rjust(self.date_width) + self.dt_separator + self.time_format,
```
Then, in `format_datetime`:

```python
if dt.date() in self.special_dates
    rv = dt.format(self.special_dates[date.date()])
```

untitaker: This would limit our capability to add special strings for times in the future ("in one hour" comes to mind)

hobarrera: "In one hour" is a datetime format, not a time format. Eg: you'd never see 'Tomorrow in one hour". So this would actually make that **simpler**.

untitaker: I just noticed that for "in one hour" we'd need an if-else chain, not just a mapping. In either case it's updated now.

todoman/ui.py
```diff
@@ -200,22 +200,47 @@ class TodoFormatter:
...
+            (None, (self.empty_date, self.empty_time)),
+        ])
+
         self._parsedatetime_calendar = parsedatetime.Calendar()
 
+    def _prepare_special_datetimes(self, mapping):
+        # Map special dates to the special string we need to return
+        self.special_datetimes = {}
+        for key, value in mapping:
+            if isinstance(value, tuple):
```
hobarrera: Is this `if` ever false?

todoman/ui.py
```diff
@@ -258,41 +283,39 @@ def detailed(self, todo):
...
-                rv = self.special_dates[date.date()]
-            else:
-                rv = date.strftime(self.date_format)
-            return rv
+    def format_datetime(self, dt):
+        dt = dt or None
+        if dt in self.special_datetimes:
+            format_string = self.special_datetimes[dt]
+        elif dt.date() in self.special_datetimes:
+            format_string = self.special_datetimes[dt.date()]
```
hobarrera: Won't this exclude the time?

todoman/ui.py
```diff
@@ -258,41 +283,39 @@ def detailed(self, todo):
...
         else:
-            return self.empty_date
+            format_string = self.datetime_format
+        if not dt:
+            return format_string
```
hobarrera: Should this be empty string?

todoman/ui.py
```diff
@@ -258,41 +283,39 @@ def detailed(self, todo):
...
-            rv = datetime.strptime(date, self.date_format)
-        except ValueError:
-            rv, certainty = self._parsedatetime_calendar.parse(date)
-            if not certainty:
-                raise ValueError(
-                    'Time description not recognized: {}' .format(date)
-                )
-            rv = datetime.fromtimestamp(mktime(rv))
-
+        rv = self._parse_datetime_naive(dt)
```
hobarrera: Maybe we can merge these functions now that they'd so much  shorter?

todoman/ui.py
untitaker: No, as you asked me to do, the format strings now include the time as well. E.g. there's a value like `"Today %H:%M"` in that dictionary (assuming a particular time_format).

todoman/ui.py
untitaker: That'd miss the `" "`-padding such that the columns are aligned properly. It is `self.special_datetimes[None]`.

todoman/ui.py
untitaker: `_parse_datetime_naive` has multiple return paths, those would all have to add the tzinfo. That's the only reason for the split. (BTW the parsing functions didn't change in my recent commits, they've been like that from the beginning)

todoman/ui.py
untitaker: Fixed. Left-over when I was trying to add `"in one hour"` (but as said that would be only implementable as `if-else`).

untitaker:
re https://github.com/pimutils/todoman/pull/117/commits/ab7a9958892595d580503332af58b6f8b2ec8fa6: Yeah, that's probably cleaner.

todoman/ui.py
hobarrera: Oh, I completely misread this, yes, it makes sense.

I don't think the first bit of the `if` is neccessary though, since we never format mere `date`s.

todoman/ui.py
hobarrera: :+1:

hobarrera:
> re ab7a995: Yeah, that's probably cleaner.

Great, I just though it was faster to push it than explain it. 😄

untitaker:
Ok i think this is ready to merge

tests/test_formatter.py
```diff
@@ -4,18 +4,20 @@
...
+    tomorrow_12am = datetime(tomorrow.year, tomorrow.month, tomorrow.day,
+                             12, 0)
     any_day = today + timedelta(days=randrange(2, 8))
 
-    assert formatter.format_date("") == "        "
-    assert formatter.format_date(today) == "   Today"
-    assert formatter.format_date(tomorrow) == "Tomorrow"
-    assert formatter.format_date(any_day) == any_day.strftime(DATE_FORMAT)
+    assert formatter.format_datetime("") == "              "
+    assert "Today" in formatter.format_datetime(today.date())
```
hobarrera: I prefer the version with the string with spaces (eg: `"    Today"`), because it also checks that we don't break column width.

todoman/ui.py
```diff
@@ -39,13 +39,13 @@ def __init__(self, todo, lists, formatter):
 
         if todo.due:
             # TODO: use proper date_format
```
hobarrera: I *think* this `TODO` is old an obsolete?

untitaker:
Since you don't appear to be on IRC:

>I would rather revert back to the old variant. Eventually I want to add support for bare dates for DUE (without time component) and this preprocessing makes it very hard.

hobarrera:
Yeah, I jump in and out of IRC, TBH.

So, does any of the rest of the stack (rfc/icalendar/`Todo`/cache) support bare `date`s?

untitaker:
No, it would all have to be refactored for that. My point is that this code will require more changes with your proposed implementation.

hobarrera:
You're right, this won't work formatting separately, sadly. :(

I wonder if we might just as well using something like [`arrow.humanize`](http://crsmithdev.com/arrow/#humanize) ([code](https://github.com/crsmithdev/arrow/blob/a78bb2ffde8582d648d54ecc0a6b04d61408e4cb/arrow/arrow.py#L703)) to do the work for us.

hobarrera:
I've also had my eye on [maya](https://github.com/kennethreitz/maya) for some time now, and it might also be capable of doing some parsing (though we  really _should_ review it before changing anything).

untitaker:
Possibly in the future, but that's a lot of code there. I think we should be more conservative about "humanizing" datetimes, preferrably without loss of precision

hobarrera:
Back to this PR though:
* Yes, my proposal is bad.
* We should also support things like `--due 10:30`. If the event has a date, only change the time. If it doesn't make it today at 10:30.

untitaker:
I force-pushed the old variant.

untitaker:
>If the event has a date, only change the time.

This will be complicated. Right now only datetimes are passed around.

hobarrera:
Sure, let's keep this simple and discuss more improvements in a separate issue.

untitaker:
I pushed a commit that allows the user to enter `--due 12:00`, and it will be parsed as "today 12am" (regardless of previous due value)

hobarrera:
> If the event has a date, only change the time.

I meant, "if it has a `datetime`" (as opposed to having `None`).

untitaker:
So to be clear, if I set `todo.due` to a `time`, you want that only the time component gets updated, not the entire value replaced? This sounds counterintuitive.

hobarrera:
My idea was that, basically, when a user sets a time (note: this is really pseuocode, not python):

```python
if todo.due:
   todo.due.set_time(time):
else:
   todo.due = datetime(today, time)
```
Does this seem too unintuitive (I'm biased because it's my idea)?

untitaker:
In the CLI module `todo` properties are set in a much more generic way.

hobarrera:
Okay then, I won't distract any more, and let's aim to keep this simple. ^_^

untitaker:
I added a bit of validation to the config parsing code, please review again.

todoman/configuration.py
```diff
@@ -83,4 +83,18 @@ def load_config():
                 'Bad {} setting, {}'.format(key, error.args[0])
             )
 
+    date_format = config['main']['date_format']
```
hobarrera: I'd rather see new validators that include this in this block (eg: like `validate_colour` and `validate_cache_path`.
Mostly for separation of concerns, and to avoid this function growing too much.

untitaker: Done!

hobarrera:
I pushed the changes I was suggesting for the validation.

untitaker:
Oops, sorry I didn't notice that there are already ConfigObj validators, or that ConfigObj even had that capability.

hobarrera:
LGTM, but since **I** pushed the last change, I'll also wait for you thumbs-up.

untitaker:
I woudl but I cannot approve my own RP

hobarrera:
Feel free to merge 🚀

untitaker:
Travis triggers a `push`-type build, but that is nonsensical since the PR branch is outside of the main repo. Since this is a required status check I can't merge the PR

hobarrera:
I squash-pushed this, travis should stop being dumb (I think it got confused when I pushed to `pimutils/time-support`, but that's still a bug on their side.

Issue #18, opened 2016-04-17
Filter by timerange


hobarrera:
This issue was moved from the issue tracker at GitLab.

---

`todo list` should be able to filter a timerange, i.e.:

```
todo list --start="before 2015-10-12"
todo list --due="before 2015-10-16"
```

like thunderbird lightning, pre-configured timerange filters would be handy:

```
todo list --timerange=current
todo list --timerange=today
todo list --timerange=7days
todo list --timerange=overdue
todo list --timerange=not_started
```

i don't know if it's better to implement a special timerange filter, or a more generic filter that can be preconfigured, which could filter by other things as well like lighning:

```
todo list filter=completed
todo list filter=incomplete
```

customizing those filters would be supreme :D

Swati4star:
Hi @hobarrera I would like to work on this issue.
Any suggestions?

untitaker:
My two cents: Implement `--start` and `--due` and see where this goes. Specialized filters can more easily be implemented as shell aliases, wrapper scripts etc, though we might want to add direct support on this.

hobarrera:
Do keep in mind that we already have a `--due` filter defined:

    --due INTEGER             Only show tasks due in DUE hours

Swati4star:
@hobarrera @untitaker I wanted to ask what exactly do you want in this
```

todo list --timerange=current
todo list --timerange=today
todo list --timerange=7days
todo list --timerange=overdue
todo list --timerange=not_started
```

It is kind of not clear. What exactly does timerange specify?

hobarrera:
I think we've covered this with `--due`. I'll revise this in future, but I think this issue is pretty much covered.

Swati4star:
Should we close this issue then?
All the mentioned features are covered now.

untitaker:
Yup

PR #121, opened 2017-02-22
add start option to todo list #18

Swati4star:
Adding start option to todo list 
For example,
`todo list --start="before 2017-03-22`

One enhancement as mentioned in https://github.com/pimutils/todoman/issues/18

@untitaker @hobarrera Please review this.


untitaker:
When setting up todoman, the user specifies a custom datetime format in their `todoman.conf` that may look entirely different from `%Y-%m-%d`. This is why we need to parse the date using `_validate_date_param`, and pass a `datetime` object along. This `datetime` may contain hours and minutes. So when passing the data to sqlite, I don't think we can use `DATE()`, but rather have to convert the `datetime` object to a timestamp and let sqlite compare integers instead of dates.

Note that the user-defined formatting string may not have the length `10`, so you will need to find a different way to slice the strings up.

Another issue here is that `--due` and `--start` now behave completely differently (sorry for not making our expectations not clear from the beginning here!). It would be nice if they accept both the same values. I don't know how that behavior should look like, perhaps @hobarrera has stronger opinions. If not, I'd say you're free to choose a behavior that you think is handy.

todoman/cli.py
```diff
@@ -331,7 +349,21 @@ def list(
...
+    before = None
+
+    # if start:
+    #     if 'before' in start:
+    #         before = True
+    #     else:
+    #         before = False
+    #     start = start[-10:]
+
+    if start:
```
untitaker: I'd prefer if we pass `start` unmodified into `db.todos` and disassemble the list there. I.e. no additional `before` parameter, just passing a list.

todoman/cli.py
```diff
@@ -332,6 +351,13 @@ def list(
 
     sort = sort.split(',') if sort else None
 
+    if start:
```
untitaker: I'd prefer if we pass the list directly into `db.todos()`. So no extra `before` parameter.

todoman/cli.py
```diff
@@ -48,6 +48,23 @@ def _validate_date_param(ctx, param, val):
         raise click.BadParameter(e)
 
 
+def _validate_start_date_param(ctx, param, val):
+    try:
+        if 'before' in val:
```
untitaker: I'd prefer `val.startswith('before ')`, because you're going to have a very weird error message when somebody enters something like `2017-09-09 before`.

todoman/cli.py
```diff
@@ -48,6 +48,23 @@ def _validate_date_param(ctx, param, val):
...
+            temp = val[7:]
+            ret = [True]
+        elif 'after' in val:
+            temp = val[6:]
+            ret = [False]
+        else:
+            raise click.BadParameter(
+                'The start date of the task should be in format \'before 2012-10-16\'')
+        ret.append(ctx.obj['formatter'].parse_date(temp))
+        return ret
```
untitaker: Instead of a list I'd prefer a tuple. Because tuples are immutable you'll have to create it in one step. Something like this:

    if ..:
        is_before = True
    elif ..:
        is_before = False
    else:
        ..
    return is_before, ctx.obj['formatter'].parse_datetime(temp)

Also please use a better variable name than `temp`.

todoman/cli.py
```diff
@@ -48,6 +48,23 @@ def _validate_date_param(ctx, param, val):
         raise click.BadParameter(e)
 
 
+def _validate_start_date_param(ctx, param, val):
+    try:
```
untitaker: Only the invocation of `parse_date` has to be in the try-except block. Moving the rest out of it spares unnecessary indentation.

todoman/cli.py
```diff
@@ -48,6 +48,23 @@ def _validate_date_param(ctx, param, val):
...
+        if 'before' in val:
+            temp = val[7:]
+            ret = [True]
+        elif 'after' in val:
+            temp = val[6:]
+            ret = [False]
+        else:
+            raise click.BadParameter(
+                'The start date of the task should be in format \'before 2012-10-16\'')
```
untitaker: This isn't entirely correct since the date format can differ now.

todoman/cli.py
```diff
@@ -313,8 +330,10 @@ def move(ctx, list, ids):
               'Defaults to true.')
 @click.option('--due', default=None, help='Only show tasks due in DUE hours',
               type=int)
+@click.option('--start', default=None, callback=_validate_start_date_param,
+              help='Only shows tasks after given DATE', type=str)
```
untitaker: I think `type=str` has no effect here, since we're already using a callback we can remove `type=str`.

todoman/model.py
```diff
@@ -603,6 +605,14 @@ def todos(self, all=False, lists=[], urgent=False, location='',
             max_due = (datetime.now() + timedelta(hours=due)).timestamp()
             extra_where.append('AND due IS NOT NULL AND due < ?')
             params.append(max_due)
+        if start:
+            if before:
+                extra_where.append('AND created_at <= ?')
+                params.append(start)
+            else:
+                extra_where.append('AND created_at >= ?')
+                params.append(start)
```
untitaker: Does this work? I think you need to call `start.timestamp()` but I'm not sure.

untitaker:
I realize this review was a bit rushed, feel free to ask questions if it's unclear *what* I mean, or *why* I want a particular change.

todoman/model.py
Swati4star: This is working. Should I still use start.timestamp()

todoman/cli.py
hobarrera: I'm also thinking that if we make the flag `--starts-before`, parsing might be a lot simpler, but I'm leaving this up to you.

todoman/model.py
untitaker: No, not at all. I didn't know that one can do that! I opened #125 for further investigation.

hobarrera: Yes, you need to pass a timestamp.

todoman.conf
```diff
@@ -0,0 +1,7 @@
+[main]
```
hobarrera: This file, along with the two eggs listed above should not have been committed.

Personally, I used `git add -p` to manually add chunks and make sure I'm not adding anything else accidentally.

todoman/model.py
untitaker: Are you sure? If it works I'm not opposed to keeping it...

Swati4star:
@hobarrera @untitaker Can you please review the code now?

Also where should I modify the changelog entry?
In v 2.0.0 or new features?

AUTHORS.rst
```diff
@@ -15,3 +15,4 @@ We manually track this since commit author *does not* equal code author.
 * Markus Unterwaditzer <markus@unterwaditzer.net>
 * Paweł Fertyk <pfertyk@openmailbox.org>
 * Thomas Glanzmann <thomas@glanzmann.de>
+* Swati Garg <swati4star@gmail.com>
```
hobarrera: This list is in alphabetical order (I'll add a clarification this time, since it's not at all obvious, really my bad).

todoman/model.py
```diff
@@ -567,6 +568,7 @@ def todos(self, all=False, lists=[], urgent=False, location='',
             with a ``-`` prepended will be used to sort in reverse order.
         :param bool reverse: Reverse the order of the todos after sorting.
         :param int due: Return only todos due within ``due`` hours.
+        :param start: Return only todos before/after ``start`` date
```
hobarrera: This doesn't make it very clear what this arg exactly is. I really do think we need separate `start_before` and `start_after` args.

Opinion, @untitaker ?

hobarrera:
Changelog entry should be for v2.2.0, since v2.1.0 is the latest release version.

todoman/model.py
untitaker: Probably.

untitaker:
@Swati4star most of my comments on https://github.com/pimutils/todoman/pull/121#pullrequestreview-23296268 still apply.

Swati4star:
@untitaker Is it possible to get the configuration file from tests?  I need it to get default date_format for testing the option start.
I am always getting `No configuration file found.` error.

untitaker:
Are you testing the entire `todo` command (the `--start` option), or just the `Database` class (the `start` parameter of `Database.todos`)?

If you test `--start`: If you use the `runner` fixture as done in `test_basic.py`, the configuration [from here](https://github.com/untitaker/todoman/blob/afc8f76f463c8dc86aaa21c6c31cd53b5269384a/tests/conftest.py#L22-L27) is used.

Swati4star:
I want to test the whole command
`todo list --start="before <date>"`

The thing is that, in the code, the date is parsed from config file. So, for tests too, I need to give date in that format only. Any help on how to do this?

Also, apart from adding the test, what else should I change?

untitaker:
>The thing is that, in the code, the date is parsed from config file. So, for tests too, I need to give date in that format only. Any help on how to do this?

In the code I linked the entire config is defined for all tests, i.e. you can assume that `date_format = %Y-%m-%d`.

untitaker:
>Also, apart from adding the test, what else should I change?

I think:

* changelog entry
* remove `.egg` files
* most comments from https://github.com/pimutils/todoman/pull/121#pullrequestreview-23296268 still need to be addressed

Swati4star:
I have added changelog entry, removed .egg files, and addressed all the comments from that review.

I am working on adding the testcase.

todoman/model.py
```diff
@@ -603,7 +605,13 @@ def todos(self, all=False, lists=[], urgent=False, location='',
             max_due = (datetime.now() + timedelta(hours=due)).timestamp()
             extra_where.append('AND due IS NOT NULL AND due < ?')
             params.append(max_due)
-
+        if start:
```
untitaker: Instead of indexing you can write `is_before, dt = start` here and use those variables. `is_before` is then the same as `start[0]`, and `dt` is the same as `start[1]`. This also validates the length of the list at the same time.

todoman/cli.py
```diff
@@ -48,6 +48,24 @@ def _validate_date_param(ctx, param, val):
         raise click.BadParameter(e)
 
 
+def _validate_start_date_param(ctx, param, val):
+    if val is None:
+        return val
+    if 'before' in val:
```
untitaker: Please use `val.startswith('before ')` here.

todoman/cli.py
```diff
@@ -48,6 +48,24 @@ def _validate_date_param(ctx, param, val):
...
+        return_val = [True]
+    elif 'after' in val:
+        return_val = [False]
+    else:
+        raise click.BadParameter(
+            'The start date of the task should be'
+            'in format \'before <date-format>\'')
+    try:
+        return_val.append(ctx.obj['formatter'].parse_date(val))
+        return return_val
```
untitaker: Please return a tuple here, not a list.

untitaker:
BTW @Swati4star as you can see we are very picky with the code, don't let that get you down :)

CHANGELOG.rst
```diff
@@ -9,7 +9,8 @@ v2.2.0
 
 * Use the system's date format as a default.
 * Show "Today" or "Tomorrow" when due date is today/tomorrow respectively.
-
+* Add ``--start="[before|after] <dat>`` option for ``todo list``. It shows lists
```
hobarrera: "dat**e**" (there's a typo). Also, it shows _todos_ created before/after this date, not _lists_.

todoman/model.py
untitaker: On second thought I think the current API is good enough.

todoman/model.py
Swati4star: Done

todoman/model.py
Swati4star: So, do I need to change this?

tests/test_filtering.py
```diff
@@ -205,3 +205,43 @@ def test_due_naive(tmpdir, runner, create):
...
+    assert 'haha' in result.output
+    assert 'hoho' in result.output
+    result = runner.invoke(cli, ['list', '--start', 'before ' + nowMinusDay])
+    assert not result.exception
+    assert 'haha' not in result.output
+    assert 'hoho' not in result.output
+    result = runner.invoke(cli, ['list', '--start', 'after ' + nowPlusDay])
+    assert not result.exception
+    assert 'haha' not in result.output
+    assert 'hoho' not in result.output
```
Swati4star: @hobarrera @untitaker Can you please take a look at this test, and figure out what wrong am I doing here? The test is currently not passing.

todoman/model.py
untitaker: No, sorry for the noise :)

todoman/cli.py
```diff
@@ -48,6 +48,24 @@ def _validate_date_param(ctx, param, val):
         raise click.BadParameter(e)
 
 
+def _validate_start_date_param(ctx, param, val):
+    if val is None:
+        return val
+    if val.startswith('before'):
```
untitaker: Two things:

* We need to check for the string `"before "`, not `"before"` (space at the end).
* `val` isn't modified, so further down you try to pass a string like `"before 2016-09-09"` to `parse_date`. Something like this is missing: `val = val[len("before "):]`

I suspect this is why the tests fail

tests/test_filtering.py
untitaker: The two tasks you created don't have a start date. I recommend running `todo new --start ...` instead of using `create`.

todoman/cli.py
Swati4star: I just tested after doing these changes. It still gives same error.

tests/test_filtering.py
```diff
@@ -205,3 +205,43 @@ def test_due_naive(tmpdir, runner, create):
...
+
+
+def test_filtering_start(tmpdir, runner, create):
+    result = runner.invoke(cli, ['list'], catch_exceptions=False)
+    assert not result.exception
+    assert result.output == ''
+
+    today = datetime.now()
+    now = today.strftime("%Y-%m-%d")
+    nowPlusDay = (today + timedelta(days=1)).strftime("%Y-%m-%d")
```
untitaker: Please use `under_scores` instead of `camelCase`

todoman/cli.py
untitaker: Sorry, yes, this isn't why the tests fail. However, it still needs to be changed, otherwise the date parsing below will not work properly. The reason why this isn't throwing any errors is because `parse_date` is not very strict.

untitaker: I left a comment [here](https://github.com/pimutils/todoman/pull/121#discussion_r102932267) as to why the tests fail

Swati4star:
@untitaker  Done with the changes. The test `test_filtering_start` is also running fine.

untitaker:
Apart from the conflict in AUTHORS.rst this looks great!

Swati4star:
@hobarrera Can you please also review the code?

tests/test_filtering.py
```diff
@@ -205,3 +205,38 @@ def test_due_naive(tmpdir, runner, create):
...
+    assert not result.exception
+    assert result.output == ''
+
+    today = datetime.now()
+    now = today.strftime("%Y-%m-%d")
+    now_plus_day = (today + timedelta(days=1)).strftime("%Y-%m-%d")
+    now_minus_day = (today + timedelta(days=-1)).strftime("%Y-%m-%d")
+    result = runner.invoke(cli, ['list', '--start', 'before ' + now])
+    assert not result.exception
+    assert result.output == ''
```
hobarrera: The filtering code returns todos where `due <= now` (less or **equal** than), so this assertion should return one entry.

tests/test_filtering.py
```diff
@@ -205,3 +205,38 @@ def test_due_naive(tmpdir, runner, create):
...
+
+    result = runner.invoke(cli, ['list', '--start', 'after ' + now])
+    assert not result.exception
+    assert result.output == ''
```
hobarrera: This one is okay since a few miliseconds will have passed.

tests/test_filtering.py
```diff
@@ -205,3 +205,38 @@ def test_due_naive(tmpdir, runner, create):
...
+    runner.invoke(cli, ['new', '-l', 'list_one', 'haha'])
+    runner.invoke(cli, ['new', '-l', 'list_one', 'hoho'])
+
+    result = runner.invoke(cli, ['list', '--start', 'after ' + now_minus_day])
+    assert not result.exception
+    assert 'haha' in result.output
+    assert 'hoho' in result.output
+    result = runner.invoke(cli, ['list', '--start', 'before ' + now_minus_day])
+    assert not result.exception
+    assert 'haha' not in result.output
```
hobarrera: For this case, better do `assert not result.output.strip()`, to make sure we're not printing **anything**. (and remove these two assertions).

tests/test_filtering.py
```diff
@@ -205,3 +205,38 @@ def test_due_naive(tmpdir, runner, create):
...
+    assert 'hoho' not in result.output
+    result = runner.invoke(cli, ['list', '--start', 'after ' + now_plus_day])
+    assert not result.exception
+    assert 'haha' not in result.output
```
hobarrera: Likewise.

tests/test_filtering.py
Swati4star: We have not created any todo before this line, so why would there be single entry?

untitaker: Yeah I'm asking myself the same thing 😁

hobarrera: Oh, sorry, I misread 😅

Swati4star:
@hobarrera Done with all the changes you mentioned.

todoman/cli.py
```diff
@@ -48,6 +48,26 @@ def _validate_date_param(ctx, param, val):
...
+    if val.startswith('before '):
+        is_before = True
+        val = val[len('before '):]
+    elif val.startswith('after '):
+        is_before = False
+        val = val[len('after '):]
+    else:
+        raise click.BadParameter(
+            'The start date of the task should be'
+            'in format \'before <date-format>\'')
```
hobarrera: `'in format "[before|after] <date>"'`. Note that you're not mentioning `after`, and it's simpler ("prettier") to user `"` inside strings.

todoman/cli.py
```diff
@@ -48,6 +48,26 @@ def _validate_date_param(ctx, param, val):
...
+    try:
+        dt = ctx.obj['formatter'].parse_date(val)
+        return (is_before, dt)
```
hobarrera: Parenthesis here are not required.

Swati4star:
Done with both @hobarrera

todoman/cli.py
```diff
@@ -317,8 +337,11 @@ def move(ctx, list, ids):
               'Defaults to true.')
 @click.option('--due', default=None, help='Only show tasks due in DUE hours',
               type=int)
+@click.option('--start', default=None, callback=_validate_start_date_param,
+              help='Only shows tasks after given DATE')
```
hobarrera: Actually, this doc isn't quite right; it can be before/after a given date. This slipped by before, sorry.

Swati4star: No problem. 
Done @hobarrera

tests/test_filtering.py
```diff
@@ -205,3 +205,36 @@ def test_due_naive(tmpdir, runner, create):
     assert len(todos) == 2
     assert todos[0].summary == "23"
     assert todos[1].summary == "1"
+
+
+def test_filtering_start(tmpdir, runner, create):
+    result = runner.invoke(cli, ['list'], catch_exceptions=False)
+    assert not result.exception
+    assert result.output == ''
```
hobarrera: This test is failing, please do `assert not result.output.strip()` instead.

untitaker:
I fixed the tests since I think they only broke because of other PRs getting merged in the meantime.

Swati4star:
This PR is ready to be merged?

untitaker:
Yes, thank you so much!

hobarrera:
Thanks! 🎉

Issue #14, opened 2016-04-17
Add list selector to TodoEditor


hobarrera:
This issue was moved from the issue tracker at GitLab.

---

Now that we support multiple vdirs, we need the editor to support picking one for new todos (and possibly existing ones too).

AnubhaAgrawal:
Hi, I am new I would like to take up this issue.

In my opinion, I would have to make changes in the code [here](https://github.com/pimutils/todoman/blob/master/todoman/model.py#L456-L479).
Am I wrong?

untitaker:
Hi @AnubhaAgrawal,

Sorry but no, this file has nothing to do with that. When running the interactive editor using `todo edit ...`, we want to show something that allows the user to move the task to a different task list. Please note this is a more difficult issue.

The entire editor interface is built up in [`TodoEditor.__init__`](https://github.com/pimutils/todoman/blob/master/todoman/ui.py#L27). The outermost widget is created on line 89, which chains all previous widgets (text fields, buttons) together.

1. We would have to create a new variable `self.current_list` which is initialized with `todo.list`.

2. [`urwid.RadioButton`](http://urwid.org/reference/widget.html#urwid.RadioButton) can do this, so you would have to create one radiobutton for each value in `self.lists`. If the radio button changes, `self.current_list` is changed to the new value

3. In `_save_inner`, `todo.list` is set to `self.current_list`

hobarrera:
_Looks like @untitaker beat me and replied while I was still typing. Still, here goes._

Hi there! Thanks for your interest in todoman.

That's actually not right. Briefly:
* `cli.py` contains all the code involved in parsing commands/options and processing that. This is the starting point when running the app (and, admittedly, the more complex part, IMHO).
* `models.py` contains all the code involved in reading/writing/processing data (todos, lists, etc).
* `ui.py` has all the code for the *U*ser *I*nterface, including the `TodoEditor` class.

This issue is, _mostly_ in `ui.py`, and the above comments cover most of the relevant details.

AnubhaAgrawal:
ok. I think firstly I'll try for some other issue

AnubhaAgrawal:
sorry, for negative comment now I want to work on this issue.
can anyone tell that if  I want to print the values in self.lists then how to I get them?

untitaker:
Hi, sorry for the slow response, if you want to find a different issue to work on that's fine of course.

[...]

PR #131, opened 2017-02-23
Add list selector to TodoEditor(for #14)

AnubhaAgrawal:
Now, I think it's correct.


untitaker:
Hello @AnubhaAgrawal, your editor is configured to use tabs instead of spaces, and it appears that this reformatted the entire file. This makes review very hard because the diff becomes very large.

If you run `tox -eflake8`, it should spit out a list of formatting problems with your code. Please fix those problems.

hobarrera:
Looks like your editor reformatted the entire code ([see here](https://github.com/pimutils/todoman/pull/131/files)).
Not only will this custom code format fail style checks, but it also makes reading the changes very complex. :(

todoman/ui.py
```diff
@@ -28,15 +28,18 @@ def __init__(self, todo, lists, formatter):
...
+        self.current_list = todo.list
         self.todo = todo
         self.lists = lists
         self.formatter = formatter
         self.saved = EditState.none
         self._loop = None
 
         self._msg_text = urwid.Text('')
-
+        urwid.RadioButton(self.lists, "label", state='first True',
```
untitaker: * You are creating a RadioButton and discard it. Instead, you need to add it to the `pile_items` list or `items`, depending on which position you want the list selector to be. In *which position of the list* you want to add the `RadioButton` also depends on where you want the list selector to be.
* You will need multiple RadioButtons, one for each list to select.
* The first argument has to be a list to which the RadioButton will append itself to. `self.lists` cannot be that list. It should be `items` or `pile_items`, but since that list is only created further down, you will need to move the entire block down.
* `on_state_change` has to be a function that sets `self.current_list` to the list the RadioButton stands for. Something like this:

      def change_current_list(new_list):
          self.current_list = new_list

      for _list in self.lists:
          RadioButton(..., on_state_change=change_current_list, user_data=_list)

todoman/ui.py
```diff
@@ -220,7 +224,6 @@ def compact(self, todo):
         """
         Returns a brief representation of a task, suitable for displaying
         on-per-line.
-
```
untitaker: Removing these lines is not necessary.

todoman/ui.py
```diff
@@ -140,12 +143,13 @@ def _save(self, btn=None):
...
 
     def _save_inner(self):
+        self.todo.list = self.current_list
         self.todo.summary = self.summary
         self.todo.description = self.description
         self.todo.location = self.location
         self.todo.due = self.formatter.parse_date(self.due)
         self.todo.start = self.formatter.parse_date(self.dtstart)
-
+
```
untitaker: This was probably your editor, please also undo this change (added whitespace)

todoman/ui.py
```diff
@@ -140,12 +143,19 @@ def _save(self, btn=None):
             raise urwid.ExitMainLoop()
 
     def _save_inner(self):
+        self.todo.list = self.current_list
         self.todo.summary = self.summary
         self.todo.description = self.description
         self.todo.location = self.location
+<<<<<<< HEAD
```
untitaker: It appears you tried to resolve the conflict, but you need to remove one of the two "versions" manually (otherwise git could do it automatically). Please remove these added lines.

todoman/ui.py
```diff
@@ -140,12 +143,19 @@ def _save(self, btn=None):
...
+        self.todo.due = self.formatter.parse_date(self.due)
+        self.todo.start = self.formatter.parse_date(self.dtstart)
+        
+=======
         self.todo.due = self.formatter.parse_datetime(self.due)
         self.todo.start = self.formatter.parse_datetime(self.dtstart)
 
+>>>>>>> 1bcdd38a08f1daa2cac75d15c1b31b439eb06497
```
untitaker: Same thing here, this is not valid Python :)

todoman/ui.py
```diff
@@ -271,8 +279,12 @@ def _format_date(self, date):
         * if date == today or tomorrow: "Today" or "Tomorrow"
         * else: return a string representing that date
         * if no date is supplied, it returns empty_date
+<<<<<<< HEAD
```
untitaker: Same thing here.

AnubhaAgrawal:
Sorry for these commits. I commit on my repository but I cloned it from your repository so these commits will go to your repository.

untitaker:
It's no problem, we will squash the entire commit history anyway.

untitaker:
Continuing at #134

PR #133, opened 2017-02-23
Use tabulate

### LINKED ISSUE #130: Use tabluate?
Since formatting is starting to get more complex (#128 would add non-predictable column widths), maybe it makes sense to use something that *actually* renders tables, rather than just tabulating with a pre-defined amount of spaces.

* tabulate might be interesting: https://bitbucket.org/astanin/python-tabulate
* How does this affect performance? How much overhead does it add?
* Does it really make the code simpler?

hobarrera:
Some testing (batches of 100 runs of `todo list`) shows that there's no real difference between this and the previous code (eg: no regressions).

The large upside is that we don't need to worry about dealing with whitespaces and aligning columns, etc in our formatter. Date/time formatting has become simpler, and its maintenance should be a lot less painful.

Also, if no todos have dates, that column will be skipped entirely.

There are some spacing changes which I'd like to avoid (I'll check a bit if they're possible to avoid or not). The changes to the tests make this obvious.

Finally, an extra empty line will get printed if there are not `todos`, which is a bit unfriendly -- though I'm thinking about replacing that with another message altogether (clarifying that you have no todos).

Closes #130

### DIFF
--- a/todoman/ui.py
+++ b/todoman/ui.py
@@ -6,6 +6,8 @@ import click
 import parsedatetime
 import urwid
 from dateutil.tz import tzlocal
+from tabulate import tabulate
+
 
 from . import widgets
 
@@ -195,11 +197,6 @@ class TodoEditor:
 
 class TodoFormatter:
 
-    # This one looks good with [X]
-    compact_format = \
-        "{id:3d} [{completed}] {urgent} {due} {summary} {list}{percent}"
-    # compact_format = "{completed} {urgent}  {due}  {summary}"
-
     def __init__(self, date_format, time_format, dt_separator):
         self.date_format = date_format
         self.time_format = time_format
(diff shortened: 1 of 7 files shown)


untitaker:
Not sure if this is worth it. I mean, we're not saving that many lines of code.

hobarrera:
We save more complexity than number of lines really.

* Variable-width columns are now possible.
* `date_formats` narrower than 8 digits won't break alignment (currently `tomorrow` overflows).
*Two  issues open issues don't have to dance around adjusting column widths (I expect more like this to come).

I wonder if anyone would care to be able to configure the `tablefmt` 🤷‍♀️

todoman/ui.py
```diff
@@ -210,67 +209,62 @@ def __init__(self, date_format, time_format, dt_separator):
...
+            if percent:
+                percent = " ({}%)".format(percent)
+            urgent = " " if todo.priority in [None, 0] else "!"
+
+            due = self.format_datetime(todo.due)
+            if todo.due and todo.due <= self.now and not todo.is_completed:
+                due = click.style(due, fg='red')
+
+            table.append([
+                "{:3d}".format(todo.id),
```
untitaker: Is the formatting here necessary?

todoman/ui.py
```diff
@@ -281,13 +275,13 @@ def _format_date(self, date):
...
         else:
-            return self.empty_date
+            return self.empty
 
     def _format_time(self, time):
         if time:
             return time.strftime(self.time_format)
         else:
-            return self.empty_time
+            return self.empty
```
untitaker: That's just an empty string, why do we have a variable for that?

todoman/ui.py
```diff
@@ -343,21 +337,23 @@ def format_database(self, database):
...
+            data = dict(
+                completed=todo.is_completed,
+                due=self.format_datetime(todo.due),
+                id=todo.id,
+                list=todo.list.name,
+                percent=todo.percent_complete,
+                summary=todo.summary,
+                # XXX: Move this into Todo itself and dedupe it
+                urgent=todo.priority not in [None, 0],
+            )
```
untitaker: The `compact` method is gone. Either we need to restore it or remove it from the other formatter.

untitaker:
>Variable-width columns are now possible.

I'm not sure why I would want to change the column width?

todoman/ui.py
hobarrera: Nope. :P

todoman/ui.py
hobarrera: Restored.

hobarrera: No, the compact method is still there, right above `compact_multiple`.

untitaker: I am talking about PorcelainFormatter.

todoman/ui.py
hobarrera: It doesn't any more. :)

todoman/ui.py
hobarrera: Removed the formating. This should be a plus to you, since you have one-digit tasks. :D

todoman/ui.py
hobarrera: Yeah, sorry, I had typed the comment, then realized what you  meant, fixed it, but still posted the comment.

hobarrera:
> I'm not sure why I would want to change the column width?

I meant the column widths can vary without us having to jump around editing tests and tweaking whitespace (as we did with `date_format`).

untitaker:
I wonder though, do we even need `compact` anymore?

hobarrera:
It's used by `todo new` and another CLI, IIRC.

untitaker:
Ah, fair enough

On 24 February 2017 14:14:17 CET, Hugo Osvaldo Barrera <notifications@github.com> wrote:
>It's used by `todo new` and another CLI, IIRC.
>
>-- 
>You are receiving this because you commented.
>https://github.com/pimutils/todoman/pull/133#issuecomment-282288116

-- 

PR #135, opened 2017-02-24
Increment sequence number upon addition

Sakshisaraswat:
I create dictionary self.todo having key sequence whose initial value is 0 and increment by 1 and saved when file is revised.

### DIFF
--- a/todoman/model.py
+++ b/todoman/model.py
@@ -322,7 +322,9 @@ class FileTodo(Todo):
         list_ = list_ or self.list
         path = os.path.join(list_.path, self.filename)
         assert path.startswith(list_.path)
-
+        self.todo = {'SEQUENCE':0}
+        y = self.todo.get('SEQUENCE')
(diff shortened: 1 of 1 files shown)


todoman/model.py
```diff
@@ -322,7 +322,9 @@ def save(self, list_=None):
         list_ = list_ or self.list
         path = os.path.join(list_.path, self.filename)
         assert path.startswith(list_.path)
-
+        self.todo = {'SEQUENCE':0}
+        y = self.todo.get('SEQUENCE')
```
untitaker: We already discussed this on IRC, but for further reference, `y` needs to be checked for being `None`, if yes, it is replaced with `0`. Two ways to do this:

    y = self.todo.get('SEQUENCE')
    if y is None:
        y = 0

or:

    y = self.todo.get('SEQUENCE', 0)

the first line with the dictionary is not necessary then

untitaker:
continued at #138

PR #136, opened 2017-02-24
Priority filtering

### LINKED ISSUE #19: Filter for priority
This issue was moved from the issue tracker at GitLab.

---

`todo list` should be able to filter by priority, i.e.:

`todo list --priority=high,medium`

or 

`todo list --priority=!low` 

---

`todo list --urgent` partially covers this. Further support requires supporting more granular priorities.

mansimarkaur:
tries to resolve #19

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,10 +4,15 @@ Changelog
 This file contains a brief summary of new features and dependency changes or
 releases, in reverse chronological order.
 
+v3.0.0
+------
+* Dropped ``--urgent`` and introduced ``--priority``
+* Added support for filtering by priority
+
 v2.2.0
 ------
 
-* Added support for filtering by priority
+* New PyPI dependency ``tabulate``.
 * Basic support for times in due dates, new ``time_format`` configuration
   parameter.
 * Use the system's date format as a default.
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,10 +9,6 @@ v3.0.0
 
 * Dropped ``--urgent`` and introduced ``--priority``
 * Added support for filtering by priority
-
-v2.2.0
-------
-
 * New PyPI dependency ``tabulate``.
 * Basic support for times in due dates, new ``time_format`` configuration
   parameter.
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,7 +7,7 @@ releases, in reverse chronological order.
 v2.2.0
 ------
 
-* Added the Granular priority support
+* Added support for filtering by priority
 * Basic support for times in due dates, new ``time_format`` configuration
   parameter.
 * Use the system's date format as a default.
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,9 @@ v3.0.0
   parameter.
 * Use the system's date format as a default.
 * Show "Today" or "Tomorrow" when due date is today/tomorrow respectively.
+* Add ``--start="[before|after] <date>`` option for ``todo list``. It shows todos
+  created before/after given date.
+* Add flag "--done-only" to todo list. Displays only completed tasks.
 
 v2.1.0
 ------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,7 @@ releases, in reverse chronological order.
 v2.2.0
 ------
 
+* Added the Granular priority support
 * Basic support for times in due dates, new ``time_format`` configuration
   parameter.
 * Use the system's date format as a default.
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,7 @@ releases, in reverse chronological order.
 
 v3.0.0
 ------
+
 * Dropped ``--urgent`` and introduced ``--priority``
 * Added support for filtering by priority
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,8 @@ releases, in reverse chronological order.
 v2.2.0
 ------
 
+* Basic support for times in due dates, new ``time_format`` configuration
+  parameter.
 * Use the system's date format as a default.
 * Show "Today" or "Tomorrow" when due date is today/tomorrow respectively.
 
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -15,7 +15,7 @@ with a semi-permanent unique id::
 
     1 [ ] !!! 2015-04-30 Close bank account (0%)
     2 [ ] !              Send minipimer back for warranty replacement (0%)
-    3 [X]     2015-03-29 Buy soy milk (100%)
+    3 [X] 2015-03-29 Buy soy milk (100%)
     4 [ ] !!             Fix the iPad's screen (0%)
     5 [ ] !!             Fix the Touchad battery (0%)
 
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -13,17 +13,18 @@ First of all, the classic usage output:
 The default action is ``list``, which outputs all tasks for all calendars, each
 with a semi-permanent unique id::
 
-    1 [ ] ! 2015-04-30 Close bank account (0%)
-    2 [ ] !            Send minipimer back for warranty replacement (0%)
-    3 [X]   2015-03-29 Buy soy milk (100%)
-    4 [ ]              Fix the iPad's screen (0%)
-    5 [ ]              Fix the Touchad battery (0%)
+    1 [ ] !!! 2015-04-30 Close bank account (0%)
+    2 [ ] !              Send minipimer back for warranty replacement (0%)
+    3 [X]     2015-03-29 Buy soy milk (100%)
+    4 [ ] !!             Fix the iPad's screen (0%)
+    5 [ ] !!             Fix the Touchad battery (0%)
 
 The columns, in order, are:
 
  * An id.
  * Whether the task has been completed or not.
- * An ``!`` indicating it's an urgent task.
+ * An ``!!!`` indicating high priority,``!!`` indicating medium priority,
+   ``!`` indicating low priority tasks.
  * The due date
  * The task summary
  * The completed percentage
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -14,7 +14,7 @@ The default action is ``list``, which outputs all tasks for all calendars, each
 with a semi-permanent unique id::
 
     1 [ ] !!! 2015-04-30 Close bank account (0%)
-    2 [ ] !              Send minipimer back for warranty replacement (0%)
+    2 [ ] !				 Send minipimer back for warranty replacement (0%)
     3 [X] 2015-03-29 Buy soy milk (100%)
     4 [ ] !!             Fix the iPad's screen (0%)
     5 [ ] !!             Fix the Touchad battery (0%)
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -372,10 +372,10 @@ def test_color_due_dates(tmpdir, runner, create, hours):
     create(
         'test.ics',
         'SUMMARY:aaa\n'
+        'PRIORITY:0\n'
         'STATUS:IN-PROGRESS\n'
         'DUE;VALUE=DATE-TIME;TZID=ART:{}\n'
         .format(due.strftime('%Y%m%dT%H%M%S'))
-        'PRIORITY:0\n'
     )
 
     result = runner.invoke(cli, ['--color', 'always'])
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -19,7 +19,6 @@ def test_basic(tmpdir, runner, create):
     create(
         'test.ics',
         'SUMMARY:harhar\n'
-        'PRIORITY:0\n'
     )
     result = runner.invoke(cli, ['list'])
     assert not result.exception
@@ -43,7 +42,6 @@ def test_no_extra_whitespace(tmpdir, runner, create):
     create(
         'test.ics',
         'SUMMARY:harhar\n'
-        'PRIORITY:0\n'
     )
     result = runner.invoke(cli, ['list'])
     assert not result.exception
@@ -55,7 +53,6 @@ def test_percent(tmpdir, runner, create):
         'test.ics',
         'SUMMARY:harhar\n'
         'PERCENT-COMPLETE:78\n'
-        'PRIORITY:0\n'
     )
     result = runner.invoke(cli, ['list'])
     assert not result.exception
@@ -73,7 +70,6 @@ def test_show_existing(tmpdir, runner, create):
         'test.ics',
         'SUMMARY:harhar\n'
         'DESCRIPTION:Lots of text. Yum!\n'
-        'PRIORITY:0\n'
     )
     result = runner.invoke(cli, ['list'])
     result = runner.invoke(cli, ['show', '1'])
@@ -86,7 +82,6 @@ def test_show_inexistant(tmpdir, runner, create):
     create(
         'test.ics',
         'SUMMARY:harhar\n'
-        'PRIORITY:0\n'
     )
     result = runner.invoke(cli, ['list'])
     result = runner.invoke(cli, ['show', '2'])
@@ -129,7 +124,6 @@ def test_default_command(tmpdir, runner, create):
     create(
         'test.ics',
         'SUMMARY:harhar\n'
-        'PRIORITY:0\n'
     )
     result = runner.invoke(cli)
     assert not result.exception
@@ -140,7 +134,6 @@ def test_delete(tmpdir, runner, create):
     create(
         'test.ics',
         'SUMMARY:harhar\n'
-        'PRIORITY:0\n'
     )
     result = runner.invoke(cli, ['list'])
     assert not result.exception
@@ -156,7 +149,6 @@ def test_copy(tmpdir, runner, create):
     create(
         'test.ics',
         'SUMMARY:test_copy\n'
-        'PRIORITY:0\n'
     )
     result = runner.invoke(cli, ['list'])
     assert not result.exception
@@ -177,7 +169,6 @@ def test_move(tmpdir, runner, create):
     create(
         'test.ics',
         'SUMMARY:test_move\n'
-        'PRIORITY:0\n'
     )
     result = runner.invoke(cli, ['list'])
     assert not result.exception
@@ -314,13 +305,11 @@ def test_sorting_output(tmpdir, runner, create):
         'test.ics',
         'SUMMARY:aaa\n'
         'DUE;VALUE=DATE-TIME;TZID=ART:20160102T000000\n'
-        'PRIORITY:0\n'
     )
     create(
         'test2.ics',
         'SUMMARY:bbb\n'
         'DUE;VALUE=DATE-TIME;TZID=ART:20160101T000000\n'
-        'PRIORITY:0\n'
     )
 
     examples = [
@@ -357,7 +346,6 @@ def test_sorting_null_values(tmpdir, runner, create):
         'test2.ics',
         'SUMMARY:bbb\n'
         'DUE;VALUE=DATE-TIME;TZID=ART:20160101T000000\n',
-        'PRIORITY:0\n',
     )
 
     result = runner.invoke(cli)
@@ -372,7 +360,6 @@ def test_color_due_dates(tmpdir, runner, create, hours):
     create(
         'test.ics',
         'SUMMARY:aaa\n'
-        'PRIORITY:0\n'
         'STATUS:IN-PROGRESS\n'
         'DUE;VALUE=DATE-TIME;TZID=ART:{}\n'
         .format(due.strftime('%Y%m%dT%H%M%S'))
@@ -395,7 +382,6 @@ def test_flush(tmpdir, runner, create):
         'test.ics',
         'SUMMARY:aaa\n'
         'STATUS:COMPLETED\n'
-        'PRIORITY:0\n'
     )
 
     result = runner.invoke(cli, ['list'])
@@ -404,7 +390,6 @@ def test_flush(tmpdir, runner, create):
     create(
         'test2.ics',
         'SUMMARY:bbb\n'
-        'PRIORITY:0\n'
     )
 
     result = runner.invoke(cli, ['list'])
@@ -417,7 +402,6 @@ def test_flush(tmpdir, runner, create):
     create(
         'test2.ics',
         'SUMMARY:bbb\n'
-        'PRIORITY:0\n'
     )
 
     result = runner.invoke(cli, ['list'])
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -314,7 +314,7 @@ def test_sorting_output(tmpdir, runner, create):
 
     examples = [
         ('-summary', ['aaa', 'bbb']),
-        ('due', ['aaa', 'bbb']),
+        ('due', ['aaa', 'bbb'])
     ]
 
     # Normal sorting, reversed by default
@@ -345,7 +345,7 @@ def test_sorting_null_values(tmpdir, runner, create):
     create(
         'test2.ics',
         'SUMMARY:bbb\n'
-        'DUE;VALUE=DATE-TIME;TZID=ART:20160101T000000\n',
+        'DUE;VALUE=DATE-TIME;TZID=ART:20160101T000000\n'
     )
 
     result = runner.invoke(cli)
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -19,6 +19,7 @@ def test_basic(tmpdir, runner, create):
     create(
         'test.ics',
         'SUMMARY:harhar\n'
+        'PRIORITY:0\n'
     )
     result = runner.invoke(cli, ['list'])
     assert not result.exception
@@ -42,6 +43,7 @@ def test_no_extra_whitespace(tmpdir, runner, create):
     create(
         'test.ics',
         'SUMMARY:harhar\n'
+        'PRIORITY:0\n'
     )
     result = runner.invoke(cli, ['list'])
     assert not result.exception
@@ -53,6 +55,7 @@ def test_percent(tmpdir, runner, create):
         'test.ics',
         'SUMMARY:harhar\n'
         'PERCENT-COMPLETE:78\n'
+        'PRIORITY:0\n'
     )
     result = runner.invoke(cli, ['list'])
     assert not result.exception
@@ -70,6 +73,7 @@ def test_show_existing(tmpdir, runner, create):
         'test.ics',
         'SUMMARY:harhar\n'
         'DESCRIPTION:Lots of text. Yum!\n'
+        'PRIORITY:0\n'
     )
     result = runner.invoke(cli, ['list'])
     result = runner.invoke(cli, ['show', '1'])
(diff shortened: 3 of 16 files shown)


todoman/model.py
```diff
@@ -585,8 +586,11 @@ def todos(self, all=False, lists=[], urgent=False, location='',
             q = ', '.join(['?'] * len(lists))
             extra_where.append('AND files.list_name IN ({})'.format(q))
             params.extend(lists)
-        if urgent:
-            extra_where.append('AND priority = 9')
+        if priority:
+            if priority == 'medium' or priority == '!low':
```
untitaker: You handle `!low` here, but it's impossible that this value is passed because it's invalid according to the validation code you have in `cli.py`.

todoman/model.py
```diff
@@ -585,8 +586,11 @@ def todos(self, all=False, lists=[], urgent=False, location='',
...
+                extra_where.append('AND priority = 5')
+            elif priority == 'high':
+                extra_where.append('AND priority = 1')
```
untitaker: `high` means anything between `1-4`, so you'll have to write `AND priority <= 4 AND priority > 0`.

For `medium` you can then use `AND priority <= 5 AND priority > 0`

For `low` you can use `AND priority <= 9 AND priority > 0`.

todoman/cli.py
```diff
@@ -336,6 +337,9 @@ def list(
 
     sort = sort.split(',') if sort else None
 
+    if priority not in ['medium', 'low', None, 'high']:
+        raise "Invalid priority"
```
untitaker: You can't raise a string. Also I'd prefer if validation of this parameter is done in an extra callback (see e.g. `_validate_list_param`.

Also I'd prefer if we translate those strings to numbers in this file, not in `model.py`.

tests/test_porcelain.py
```diff
@@ -14,7 +14,7 @@ def test_list_all(tmpdir, runner, create):
     assert (
         result.output.strip() ==
         '{"completed": true, "due": 1451689200, "id": 1, "list": "default'
-        '", "percent": 26, "summary": "Do stuff", "urgent": false}'
+        '", "percent": 26, "summary": "Do stuff", "priority": None}'
```
hobarrera: The output of `--porcelain` cannot lose any fields (it's part of the promise we keep, since other apps can integrate into it). We **can** add fields though.

todoman/cli.py
hobarrera: Additionally, I'd also prefer if this validation function returned the numeric priority based on the input string as well, so converting string -> number is done strictly in the `cli` code..

tests/test_porcelain.py
mansimarkaur: So, we can't remove urgent and will have to add priority as another field?

todoman/cli.py
mansimarkaur: I'll add a `_validate_priority_param` method to handle this validation in `cli`

todoman/model.py
mansimarkaur: Yes, I missed this point. Thanks.

tests/test_porcelain.py
hobarrera: We can't remove urgent, nope. And I think it makes sense to add priority. We **can** show the numeric value here.

todoman/model.py
mansimarkaur: How do you think the numeric priority should be returned, as in how do I associate low with a number in the range 6-9.
One way, would be to use random function.
Any other suggestions?

untitaker: I would always just choose the highest possible number, i.e. `9` for low, `4` for high. This makes this particular code easier to write.

mansimarkaur: I'll do this. Thanks.

mansimarkaur:
I pushed changes to the `ui.py` file by mistake. I'll just revert the changes.

tests/test_filtering.py
```diff
@@ -36,16 +55,51 @@ def test_priority(tmpdir, runner, create):
     create(
         'one.ics',
         'SUMMARY:haha\n'
-        'PRIORITY: 9\n'
+        'PRIORITY: high\n'
```
untitaker: I don't think you can do that. It has to be a number.

CHANGELOG.rst
```diff
@@ -7,6 +7,7 @@ releases, in reverse chronological order.
 v2.2.0
 ------
 
+* Added the Granular priority support
```
untitaker: This isn't very informative, something like "add support for filtering by priority"

tests/test_porcelain.py
```diff
@@ -23,14 +23,14 @@ def test_list_nodue(tmpdir, runner, create):
         'test.ics',
         'SUMMARY:Do stuff\n'
         'PERCENT-COMPLETE:12\n'
-        'PRIORITY:1\n'
+        'URGENT:1\n'
```
untitaker: iCalendar doesn't have a URGENT property though.

tests/test_filtering.py
```diff
@@ -36,7 +35,7 @@ def test_urgent(tmpdir, runner, create):
     create(
         'one.ics',
         'SUMMARY:haha\n'
-        'PRIORITY: 9\n'
+        'URGENT: 9\n'
```
hobarrera: Nope, this field is still called priority.

todoman/ui.py
```diff
@@ -303,6 +305,19 @@ def format_datetime(self, dt):
             self._format_time(time_part)
         )))
 
+    def parse_priority(self, priority):
+        if priority not in ['low', 'high', 'medium', 'none']:
```
untitaker: It should be sufficient to have a single if-else chain here. Also we need to raise a ValueError on invalid values, not return None.

tests/test_filtering.py
```diff
@@ -48,6 +47,61 @@ def test_urgent(tmpdir, runner, create):
...
+    )
+    create(
+        'three.ics',
+        'SUMMARY:hehe\n'
+        'PRIORITY: medium\n'
+    )
+    create(
+        'four.ics',
+        'SUMMARY:huhu\n'
+        'PRIORITY: none\n'
```
hobarrera: Should be `0`, not `none`.

todoman/ui.py
```diff
@@ -352,7 +367,7 @@ def compact(self, todo):
             percent=todo.percent_complete,
             summary=todo.summary,
             # XXX: Move this into Todo itself and dedupe it
```
untitaker: This comment can be removed

tests/test_filtering.py
```diff
@@ -48,6 +47,61 @@ def test_urgent(tmpdir, runner, create):
...
 
+def test_priority(tmpdir, runner, create):
+    result = runner.invoke(cli, ['list'], catch_exceptions=False)
+    assert not result.exception
+    assert result.output == ''
+
+    create(
+        'one.ics',
+        'SUMMARY:haha\n'
+        'PRIORITY: high\n'
```
hobarrera: Priorities in files are numeric, so this should be 1-4.

todoman/ui.py
untitaker: To be clear, something like:

    if priority == 'low':
        ...
    elif priority == 'medium':
        ...
    ...
    else:
        raise ValueError('Priority has to be one of ...')

tests/test_porcelain.py
```diff
@@ -23,7 +23,7 @@ def test_list_nodue(tmpdir, runner, create):
...
```
hobarrera: Again, `PRIORITY`.

todoman/model.py
```diff
@@ -587,6 +589,8 @@ def todos(self, all=False, lists=[], urgent=False, location='',
             params.extend(lists)
         if urgent:
             extra_where.append('AND priority = 9')
+        elif priority:
+            extra_where.append('AND priority = ' + priority)
```
hobarrera: I think this condition needs to be a bit more complex; if I filter `--priority 5`, I'd want anything 1-5 (eg: less than 5, excluding 0).

todoman/ui.py
```diff
@@ -295,6 +324,19 @@ def format_datetime(self, dt):
             self._format_time(time_part)
         )))
 
+    def parse_priority(self, priority):
+        if priority not in ['low', 'high', 'medium', 'none']:
```
hobarrera: Actually, if it's not on this list, we should fail. We might also want to accept numbers as input (I don't have a strong opinion on this though).

todoman/ui.py
```diff
@@ -295,6 +324,19 @@ def format_datetime(self, dt):
...
+            return None
+        else:
```
hobarrera: 👏  This is great!

untitaker:
Damn, now there are two reviews, one by @hobarrera and one by me. Sorry about that! I'll retract a bit.

tests/test_filtering.py
mansimarkaur: I have a doubt.
If priority = 9 implies urgent, aren't we wrong in keeping high as 4, medium as 5  and low as 9?
Shouldn't it be none 0, medium 5, low 4, high 9?

untitaker: >If priority = 9 implies urgent

Yeah that part confuses me as well. I think the tests have a bug. The iCalendar RFC clearly states that 1-4 is priority=high

todoman/model.py
mansimarkaur: @untitaker suggested that `--priority medium` would be the way to filter and then medium would be converted to the value 5 using the `validate_priority_param`.
Do we want to keep filtering using integer values as well?

untitaker:
To be clear @hobarrera we cannot remove either `--urgent` or the `urgent` JSON field?

todoman/model.py
untitaker: Yes I suppose this would be a useful (but not necessary) feature, though @hobarrera only meant that you need to use `<=` instead of `=` here.

todoman/ui.py
mansimarkaur: Maybe we could add support for numbers in another PR?

untitaker: Yes, I think so

tests/test_filtering.py
mansimarkaur: It was a bug in the tests. Changed it.

todoman/model.py
mansimarkaur: If a user wants all tasks with medium priority, we show all the tasks that are of medium and high priority. Shouldn't we be just showing the tasks at the level 5 and not level <=5?

untitaker: But you have documented `--priority` like this:

>Only show tasks with priority greater then the specified one

todoman/model.py
```diff
@@ -567,6 +567,8 @@ def todos(self, all=False, lists=[], urgent=False, location='',
             with a ``-`` prepended will be used to sort in reverse order.
         :param bool reverse: Reverse the order of the todos after sorting.
         :param int due: Return only todos due within ``due`` hours.
+        :param str priority: Only return todos with priority greater than
```
untitaker: Replace "greater than" with "at least as high as" here (`>` vs `>=`), same in the documentation of `--priority` in `cli.py`.

todoman/model.py
mansimarkaur: Yes, I did but I was just asking what would be a better approach.
I'll change the documentation accordingly then.
What do you think would be a better way?

untitaker: I think it's better to show the tasks with priority *at least* as high as the given one, not the ones with the exact given priority.

hobarrera:
> To be clear @hobarrera we cannot remove either --urgent or the urgent JSON field?

Uhm, I'm not sure what to do here. The idea is that `--porcelain` does not change, but if we drop `--urgent`, backwards compatibility is dropped anyway, so it's not so crazy to break `--porcelain`. To be honest, there are no integrations in existence yet either.

So, I'm not keen on a precedent of "we changed `--porcelain`", but it's senseless to keep it unaltered if we drop `--urgent` too.

I think it's okay to drop the `urgent` field in the json, and the flag, if you agree, @untitaker .

CHANGELOG.rst
```diff
@@ -7,6 +7,7 @@ releases, in reverse chronological order.
 v2.2.0
```
hobarrera: Since we're dropping `--urgent`, we're breaking backward compatibility, so we'd have to push this to `3.0.0`. Please update this as part of the PR.

untitaker:
Yes I do agree. I think it's too early to stabilize anything in todoman

On 24 February 2017 19:32:58 CET, Hugo Osvaldo Barrera <notifications@github.com> wrote:
>> To be clear @hobarrera we cannot remove either --urgent or the urgent
>JSON field?
>
>Uhm, I'm not sure what to do here. The idea is that `--porcelain` does
>not change, but if we drop `--urgent`, backwards compatibility is
>dropped anyway, so it's not so crazy to break `--porcelain`. To be
>honest, there are no integrations in existence yet either.
>
>So, I'm not keen on a precedent of "we changed `--porcelain`", but it's
>senseless to keep it unaltered if we drop `--urgent` too.
>
>I think it's okay to drop the `urgent` field in the json, and the flag,
>if you agree, @untitaker .
>
>-- 
>https://github.com/pimutils/todoman/pull/136#issuecomment-282367750

-- 

CHANGELOG.rst
```diff
@@ -7,6 +7,7 @@ releases, in reverse chronological order.
...
 ------
 
+* Added support for filtering by priority
```
hobarrera: Please also mention that we're dropping `--urgent` (eg: `"replace --urgent with --priority"`).

tests/test_filtering.py
```diff
@@ -36,7 +35,7 @@ def test_urgent(tmpdir, runner, create):
     create(
         'one.ics',
         'SUMMARY:haha\n'
-        'PRIORITY: 9\n'
+        'PRIORITY:4\n'
```
hobarrera: Good catch! :+1:

hobarrera:
Okay, so a few extra changes, and also feel free to strip out `--urgent`. Looks great so far! ^_^

todoman/model.py
```diff
@@ -587,6 +589,12 @@ def todos(self, all=False, lists=[], urgent=False, location='',
             params.extend(lists)
         if urgent:
             extra_where.append('AND priority = 9')
+        elif priority:
```
untitaker: I think this line has to be `elif priority is not None`, because if `priority` is `0`, this branch is supposed to be used too

CHANGELOG.rst
```diff
@@ -4,10 +4,15 @@ Changelog
...
-v2.2.0
+v3.0.0
 ------
 
+* Dropped ``--urgent`` and introduced ``--priority``
 * Added support for filtering by priority
+
+v2.2.0
+------
+
```
hobarrera: v2.2.0 will never exist, it's safe to drop these four lines.

mansimarkaur: Should I add these lines to the v3.0.0  or should I just remove them?

tests/test_porcelain.py
```diff
@@ -8,13 +8,14 @@ def test_list_all(tmpdir, runner, create):
         'STATUS:COMPLETED\n'
         'DUE;VALUE=DATE-TIME;TZID=CET:20160102T000000\n'
         'PERCENT-COMPLETE:26\n'
+        'PRIORITY:0'
```
untitaker: We need a `\n` here, otherwise the tests crash.

todoman/ui.py
```diff
@@ -224,7 +219,14 @@ def compact_multiple(self, todos):
             percent = todo.percent_complete or ''
             if percent:
                 percent = " ({}%)".format(percent)
-            urgent = " " if todo.priority in [None, 0] else "!"
+            if todo.priority == 5:
+                priority = "!!"
+            elif todo.priority == 4:
```
untitaker: Here we have to check for the entire possible range (1-4), because the task could be created by another app (Android's OpenTask, Nextcloud tasks, iCloud, Microsoft Office365, ...)

mansimarkaur: So, will we have to raise an error also, if the priority does not belong to the range 0-9?

CHANGELOG.rst
hobarrera: I meant "remove the v2.2.0" title; the changes should stay since they be released in 3.0.0

todoman/ui.py
untitaker: I thínk this sort of validation should happen somewhere else, so I think it's not necessary. I filed https://github.com/pimutils/todoman/issues/140 because we haven't thought about this.

tests/test_basic.py
```diff
@@ -70,6 +73,7 @@ def test_show_existing(tmpdir, runner, create):
         'test.ics',
         'SUMMARY:harhar\n'
         'DESCRIPTION:Lots of text. Yum!\n'
+        'PRIORITY:0\n'
```
untitaker: Apparently GitHub swallowed my comment: Adding this line to all tests shouldn't be necessary, todoman should use `0` as value if `PRIORITY` is not defined.

todoman/model.py
```diff
@@ -585,8 +586,12 @@ def todos(self, all=False, lists=[], urgent=False, location='',
...
             extra_where.append('AND files.list_name IN ({})'.format(q))
             params.extend(lists)
-        if urgent:
-            extra_where.append('AND priority = 9')
+        if priority is not None:
+            if priority == 0:
+                extra_where.append('AND priority = 0')
+            else:
+                extra_where.append('AND priority >0'
+                                   ' AND priority <= ' + priority)
```
untitaker: You are adding a string and an integer. I recommend using `"AND PRIORITY > 0 AND PRIORITY <= {}".format(priority)` instead.

tests/test_filtering.py
```diff
@@ -28,24 +28,58 @@ def test_all(tmpdir, runner, create):
...
+    result_low = runner.invoke(cli, ['list', '--priority=low'])
+    assert not result_low.exception
+    assert 'haha' in result_low.output
+    assert 'hehe' in result_low.output
+    assert 'hoho' in result_low.output
+    assert 'huhu' not in result_low.output
+
+    result_none = runner.invoke(cli, ['list', '--priority=none'])
+    assert not result_none.exception
+    assert 'haha' in result_none.output
```
untitaker: Shouldn't only `huhu` be in the output? It is the only task without priority.

todoman/model.py
hobarrera: No, actually, check how lines above use `params` to insert values into the query safely (this avoids possible issues with quoting, etc).

todoman/ui.py
```diff
@@ -343,12 +361,24 @@ def compact(self, todo):
...
             percent=todo.percent_complete,
             summary=todo.summary,
-            # XXX: Move this into Todo itself and dedupe it
-            urgent=todo.priority not in [None, 0],
+            priority=todo.priority,
         )
 
         return json.dumps(data, sort_keys=True)
 
+    def parse_priority(self, priority):
```
untitaker: The tests crash because `parse_priority` isn't defined for `PorcelainFormatter`. For this purpose I think it would be a good idea to make `PorcelainFormatter` a subclass of the normal formatter, what do you think @hobarrera?

todoman/model.py
```diff
@@ -585,8 +586,12 @@ def todos(self, all=False, lists=[], urgent=False, location='',
...
             extra_where.append('AND files.list_name IN ({})'.format(q))
             params.extend(lists)
-        if urgent:
-            extra_where.append('AND priority = 9')
+        if priority is not None:
+            if priority == 0:
+                extra_where.append('AND priority = 0')
+            else:
+                extra_where.append('AND priority >0'
+                                   ' AND priority <= ({})'.format(priority))
```
untitaker: Sorry for the incorrect correction earlier. It turns out you have to use `(?)` here and append `priority` to `params` (see line 588).

todoman/ui.py
mansimarkaur: I just defined a `parse_priority` for `PorcelainFormatter`, why doesn't that work?

untitaker: Yes, that works too. Unsure if that's the solution we want (duplicating code) though

mansimarkaur: Yeah, even I was feeling a bit weird copying it from `TodoFormatter`.
It would be better to make it a subclass.

tests/test_porcelain.py
```diff
@@ -23,14 +24,15 @@ def test_list_nodue(tmpdir, runner, create):
...
         'SUMMARY:Do stuff\n'
         'PERCENT-COMPLETE:12\n'
-        'PRIORITY:1\n'
+        'PRIORITY:4\n'
     )
 
     result = runner.invoke(cli, ['--porcelain', 'list'])
+    assert not result.exception
     assert (
-        result.output.strip() ==
```
untitaker: I'm not sure if you intentionally removed the `strip()`, but removing it makes the tests fail.

untitaker:
We discussed this on IRC, but for the record, I misunderstood how `--priority=none` is supposed to behave and therefore made some incorrect review comments. I thought it should show only tasks without priority, but it makes much more sense (as intended by @mansimarkaur) to show all tasks.

todoman/model.py
```diff
@@ -585,8 +586,13 @@ def todos(self, all=False, lists=[], urgent=False, location='',
...
             params.extend(lists)
-        if urgent:
-            extra_where.append('AND priority = 9')
+        if priority is not None:
+            if priority == 0:
+                extra_where.append('AND PRIORITY <= 9 AND PRIORITY >= ?')
+                params.append('{}'.format(priority))
+            else:
+                extra_where.append('AND PRIORITY > 0 AND PRIORITY <= ?')
+                params.append('{}'.format(priority))
```
untitaker: The entire code can be collapsed to:

    if priority:
        extra_where.append('AND PRIORITY > 0 AND PRIORITY <= ?')
        params.append('{}'.format(priority))

Because in the case of `priority == 0` basically no filtering is done, so we don't need to append anything to `extra_where`.

tests/test_basic.py
```diff
@@ -314,7 +314,7 @@ def test_sorting_output(tmpdir, runner, create):
 
     examples = [
         ('-summary', ['aaa', 'bbb']),
-        ('due', ['aaa', 'bbb'])
+        ('due', ['aaa', 'bbb']),
```
hobarrera: This change is irrelevant to this PR. There are a few other comma additions/removals, please exclude these (they just add noise to the diff/history).

todoman/ui.py
```diff
@@ -224,7 +222,14 @@ def compact_multiple(self, todos):
...
                 percent = " ({}%)".format(percent)
-            urgent = " " if todo.priority in [None, 0] else "!"
+            if todo.priority == 5:
+                priority = "!!"
+            elif todo.priority <= 4 and todo.priority >= 1:
+                priority = "!!!"
+            elif todo.priority <= 9 and todo.priority >= 6:
+                priority = "!"
+            elif todo.priority == 0:
+                priority = " "
```
hobarrera: Return an empty string here. Indentation/spacing is handled elsewhere so no need to worry about that.

hobarrera: (Also, if it's an empty string we don't add a column if no todos have priority).

todoman/ui.py
hobarrera: IMHO, the porcelain formatted should accept things like `"0"`, `"1"`, etc.

todoman/ui.py
mansimarkaur: Is this how, it should be shown then:
```
    1 [ ] !!! 2015-04-30 Close bank account (0%)
    2 [ ] !              Send minipimer back for warranty replacement (0%)
    3 [X] 2015-03-29 Buy soy milk (100%)
    4 [ ] !!             Fix the iPad's screen (0%)
    5 [ ] !!             Fix the Touchad battery (0%)
```

mansimarkaur:
So, for porcelain formatter's `parse_priority` should take "0" and "1" as strings and then parse it into numbers?

untitaker:
@mansimarkaur 

>So, for porcelain formatter's parse_priority should take "0" and "1" as strings and then parse it into numbers?

Yes I think that's what @hobarrera meant

docs/source/usage.rst
```diff
@@ -15,7 +15,7 @@ with a semi-permanent unique id::
 
     1 [ ] !!! 2015-04-30 Close bank account (0%)
     2 [ ] !              Send minipimer back for warranty replacement (0%)
-    3 [X]     2015-03-29 Buy soy milk (100%)
+    3 [X] 2015-03-29 Buy soy milk (100%)
```
untitaker: Does this actually happen? When I test todoman, it doesn't.

todoman/ui.py
```diff
@@ -150,12 +153,7 @@ def _save_inner(self):
 
         self.todo.is_completed = self._completed.get_state()
 
-        # If it was already non-zero, keep it that way. Let's not overwrite
-        # values 1 thru 8.
-        if self._urgent.get_state() and not self.todo.priority:
-            self.todo.priority = 9
-        elif not self._urgent.get_state():
-            self.todo.priority = 0
+        self.todo.priority = self.formatter.parse_priority(self._priority)
```
untitaker: This doesn't work, because `self._priority` is a ExtendedEdit, not a text by itself. You need to add a property like this further below:

    @property
    def priority(self):
        return self._priority.edit_text

and call parse_priority with `self.priority` instead of `self._priority`

todoman/ui.py
```diff
@@ -51,6 +51,9 @@ def __init__(self, todo, lists, formatter):
         else:
             dtstart = ''
 
+        if todo.priority:
```
untitaker: The `else` branch is missing here. `todo edit` crashes on tasks without priority. This should work:

    ...
    else:
        priority = ''

todoman/ui.py
```diff
@@ -373,16 +373,16 @@ def compact(self, todo):
         return json.dumps(data, sort_keys=True)
 
     def parse_priority(self, priority):
```
hobarrera: Great! Please also add some tests for this method (there's `tests_porcelain` in a separate file).

untitaker:
You left IRC so quickly! Your tests fail because in porcelain-mode, `--priority` only accepts integers, not the string `"high"`.

mansimarkaur:
@untitaker Hasn't priority's type been set as `str` here: https://github.com/mansimarkaur/todoman/blob/8c8af792c321552f9b701f2bbe3a6689cfaca261/todoman/cli.py#L347 ?

untitaker:
Ah, I didn't mean the type, but "integers" as in `"1"`, `"2"` and so on (as opposed to `"high"`). Only when using `--porcelain`

todoman/ui.py
```diff
@@ -343,12 +367,27 @@ def compact(self, todo):
...
-            urgent=todo.priority not in [None, 0],
+            priority=todo.priority,
         )
 
         return json.dumps(data, sort_keys=True)
 
+    def parse_priority(self, priority):
+        if priority is None:
+            return 0
+        prio_int = int(priority)
```
untitaker: Concretely, this line will throw an exception whenever `priority` is not an "integer"

todoman/ui.py
```diff
@@ -343,12 +367,19 @@ def compact(self, todo):
...
         )
 
         return json.dumps(data, sort_keys=True)
 
+    def parse_priority(self, priority):
+        if priority is None:
+            return 0
+        try:
+            return int(priority)
+        except:
```
untitaker: Please catch only `ValueError` here

hobarrera:
Thanks! 🎉

PR #138, opened 2017-02-24
Increment sequence number upon edition

Sakshisaraswat:
Increment sequence number upon edition.a variable y is taken and value of self.todo['SEQUENCE'] is taken in this variable . value of y is checked and if it is none then value of y is set to 0 and if not then value of sequence is incremented by one

### DIFF
--- a/todoman/model.py
+++ b/todoman/model.py
@@ -322,7 +322,7 @@ class FileTodo(Todo):
         list_ = list_ or self.list
         path = os.path.join(list_.path, self.filename)
         assert path.startswith(list_.path)
-        sequence = self.todo.get('SEQUENCE',0)
+        sequence = self.todo.get('SEQUENCE', 0)
         self.todo['SEQUENCE'] = sequence + 1
         if os.path.exists(path):
             # Update an existing entry:
--- a/todoman/model.py
+++ b/todoman/model.py
@@ -322,10 +322,10 @@ class FileTodo(Todo):
         list_ = list_ or self.list
         path = os.path.join(list_.path, self.filename)
         assert path.startswith(list_.path)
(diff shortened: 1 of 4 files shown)


untitaker:
Looks good to me! Please add a testcase in `test_model.py`, a changelog entry and yourself to AUTHORS. Thanks!

todoman/model.py
```diff
@@ -322,7 +322,10 @@ def save(self, list_=None):
         list_ = list_ or self.list
         path = os.path.join(list_.path, self.filename)
         assert path.startswith(list_.path)
-
+        y = self.todo.get('SEQUENCE')
```
hobarrera: I'd rather this variable be called `seq` or `sequence`, rather than `y` (eg: something more descriptive).

Sakshisaraswat:
ok i will make changes

Sakshisaraswat:
Please tell me how to write test cases?

todoman/model.py
```diff
@@ -322,7 +322,10 @@ def save(self, list_=None):
         list_ = list_ or self.list
         path = os.path.join(list_.path, self.filename)
         assert path.startswith(list_.path)
-
+        sequence = self.todo.get('SEQUENCE')
+        if sequence is None:
+            sequence = 0
```
hobarrera: These three lines can actually all be simplified to `sequence = self.todo.get('SEQUENCE', 0)`. What it does is return the sequence, of if it's `None`, return the second value (which is `0`).

hobarrera:
I took the liberty of fixing merge conflicts, but the prior review is still pending.

hobarrera:
Thanks! :+1:

Issue #118, opened 2017-02-22
todo does not output anything when their are no list. 


AnubhaAgrawal:
**Actual behavior**
When running for the first time or when there are no lists, the `todo` command outputs nothing. This is very counter-intuitive and may confuse the users.
**Expected behavior**
When there are no lists, a warning should be displayed saying the same and a suggestion to display help should be shown.
I would like to work on this issue. Can I send a PR?

untitaker:
This makes sense, and yes, PR is welcome!

I think all that is necessary is a check whether the path is empty, [right here](https://github.com/untitaker/todoman/blob/ba9b8c59b52d25a946d074713b88c309e2d02a5e/todoman/cli.py#L112).

Also a command to show all task lists would probably be useful for debugging.

PR #134, opened 2017-02-24
Add list selector to TodoEditor(for #14)

AnubhaAgrawal:
Sorry for the previous commit.
 Now, please check this.

PR #141, opened 2017-02-26
Nolist

AnubhaAgrawal:
todo does not output anything when their are no list.(for  #118)

### DIFF
--- a/todoman/cli.py
+++ b/todoman/cli.py
@@ -349,5 +349,4 @@ def list(
         urgent=urgent,
     )
 
-    for todo in todos:
-        click.echo(ctx.obj['formatter'].compact(todo))
+    click.echo(ctx.obj['formatter'].compact_multiple(todos))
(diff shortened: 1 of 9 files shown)


untitaker:
@AnubhaAgrawal your pull request contains changes from other pull requests, particularly your other one, #134. Please remove those changes (`git rebase -i master` works for this).

todoman/cli.py
```diff
@@ -113,6 +113,9 @@ def cli(ctx, color, porcelain):
         path for path in glob.iglob(expanduser(config["main"]["path"]))
         if isdir(path)
     ]
+    if len(paths) == 0:
+        print("list is empty\n")
```
untitaker: Please use `click.echo` instead of `print` and also call `ctx.exit(1)` such that todoman doesn't do anything afterwards. Also:

* I prefer something like "No lists found." over "list is empty".
* we should give the user more information: Where to create a new folder (for that we can print out `config["main"]["path"]`.
* `todo --help` doesn't help the user here.

untitaker:
Continuing at #142

PR #142, opened 2017-02-26
todo does not output anything when their are no list.(for #118)

AnubhaAgrawal:
todo does not output anything when their are no list.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -21,7 +21,7 @@ v3.0.0
 * Make the output of move, delete, copy and flush more consistent with
   everything else.
 * Porcelain now outputs proper JSON, rather than one-JSON-per-line.
-
+* todo does not output anything when their are no list.
 
 v2.1.0
 ------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -21,8 +21,8 @@ v3.0.0
 * Make the output of move, delete, copy and flush more consistent with
   everything else.
 * Porcelain now outputs proper JSON, rather than one-JSON-per-line.
-* Make todoman print "No lists found matching" and tell the user how to create a list when there are none.
 * Increment sequence number upon edits.
+* Print a descriptive message when no lists are found.
 
 v2.1.0
 ------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -21,7 +21,7 @@ v3.0.0
 * Make the output of move, delete, copy and flush more consistent with
   everything else.
 * Porcelain now outputs proper JSON, rather than one-JSON-per-line.
-* todo does not output anything when their are no list.
+* Make todo to print "No lists found matching" and tell How do I create a list? and Where do I create it? when their are no list.
 * Increment sequence number upon edits.
 
 v2.1.0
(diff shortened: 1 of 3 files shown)


untitaker:
@AnubhaAgrawal Thanks, this looks good. Perhaps you could add a bit more text that explains the part that you found not self-explanatory. A few questions arise: 1.) How do I create a list? 2.) Where do I create it?

todoman/cli.py
```diff
@@ -143,6 +143,11 @@ def cli(ctx, color, porcelain):
     if len(paths) == 0:
         click.echo("No lists found.\n")
         click.echo(config["main"]["path"])
+        click.echo("goto ~/.local/share/calendars")
```
untitaker: `~/.local/share/calendars` is the default path, but the user may have configured another one. This is what `config["main"]["path"]` is for.

Something like: `"No lists found matching {}, create a directory for a new list".format(config["main"]["path"]`

untitaker:
Also please add a testcase and changelog entry.

You can prepare a testcase using the `runner` fixture [as shown here](https://github.com/pimutils/todoman/blob/master/tests/test_basic.py#L14). However, by default there is already a list created, so you also need to do `tmpdir.join('default').remove()` before testing.

tests/test_basic.py
```diff
@@ -425,5 +426,17 @@ def test_edit(runner, default_database):
     assert todo.due == datetime.datetime(2017, 2, 1, tzinfo=tzlocal())
     assert todo.summary == 'Eat paint'
 
+
+def test_empty_list(tmpdir, runner, create):
+    [item.remove()
```
untitaker: Please use a normal for-loop here.

CHANGELOG.rst
```diff
@@ -21,7 +21,7 @@ v3.0.0
 * Make the output of move, delete, copy and flush more consistent with
   everything else.
 * Porcelain now outputs proper JSON, rather than one-JSON-per-line.
-
+* todo does not output anything when their are no list.
```
hobarrera: Please rewrite this. This changelog should explain in simple term to users what changed. This explains what the bug was, not what we changed.

todoman/cli.py
```diff
@@ -144,6 +144,11 @@ def cli(ctx, color, porcelain):
         path for path in glob.iglob(expanduser(config["main"]["path"]))
         if isdir(path)
     ]
+    if len(paths) == 0:
+        click.echo("No lists found matching {}, "
```
hobarrera: This string can fit into a single line (78 characters), no need to split it.

untitaker: No, it doesn't fit. The string will wrap on the last word.

hobarrera:
```
In [1]: len('    click.echo("No lists found matching {}, create a directory for a new list"')
Out[1]: 78
```

untitaker: You're missing a level of indentation.

hobarrera:
Also, not strictly related to this PR, @AnubhaAgrawal, can you check your system's time? Your commits have dates from the future (and github doesn't seem to mind about that).

AnubhaAgrawal:
Now, I correct my system's time, please tell me what should I do now?

untitaker:
LGTM

untitaker:
Thanks @AnubhaAgrawal for your PR!

PR #150, opened 2017-02-27
Don't store generator

untitaker:


hobarrera:
Exactly what impact does this have?

untitaker:
It seems wrong to leave a drained iterator lying around in the instance. We could also remove `self.lists` and dispose of the generator after we're done with it.

hobarrera:
I'd prefer to generate the list before passing it to the constructor, so this doesn't kick us back unexpectedly when we reuse it working on the full interactive interface.

untitaker:
>so this doesn't kick us back unexpectedly when we reuse it working on the full interactive interface.

Yes, this is the reason I made this PR. Why rely on the caller to fulfill contract though?

hobarrera:
> Pull requests that have a failing status can’t be merged on a phone.

untitaker:
Do you approve this PR? It appears only codecov's status fails

hobarrera:
Yes, es. I tried to merge it but Couldn't.

PR #151, opened 2017-02-27
Proper porcelain json

hobarrera:
Make tests simpler, and output proper JSON.

Since the next release will alter porcelain, let's go the whole way (also, this makes using `jq` or other stuff less painful).

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,6 +20,7 @@ v3.0.0
 * Add flag "--done-only" to todo list. Displays only completed tasks.
 * Make the output of move, delete, copy and flush more consistent with
   everything else.
+* Porcelain now outputs proper JSON, rather than one-JSON-per-line.
 
 
 v2.1.0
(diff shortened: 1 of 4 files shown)


untitaker:
you're sure jq can't handle the old format?

hobarrera:
The old format wasn't proper JSON, but "one JSON per line". Maybe `jq` could, but most `JSON` libraries wouldn't (not without extra work).

untitaker:
FWIW http://jsonlines.org/ isn't uncommon practice.

untitaker:
Also I liked the old format for being "streamable", while the new one has to be output as single chunk.

hobarrera:
Huh, I was unaware of that. So, do you prefer jsonl to json? I kinda prefer JSON because it's a bit more "standard" (especially considering apps will parse this, and never humans). But I don't have very strong opinions on it.

untitaker:
In general I think the fact that we have to wait for the entire data before we're able to format the output is a bad thing. That is also the case ẁith using the `tabulate` pkg instead of using fixed columns.

Issue #159, opened 2017-03-01
`todo --porcelain edit ...` crashes


untitaker:
When providing todoman with the porcelain option, the TodoEditor crashes. This can be fixed together with https://github.com/pimutils/todoman/issues/158 by not relying on `ctx.obj["formatter"]` at all.

hobarrera:
`PorcelainFormatter` should have the same API as `TodoFormatter`, so if it's lacking any methods, or methods have different signatures, it's a bug and I'm okay with changing it. :+1:

PR #160, opened 2017-03-01
Introduce AppContext, lazy init for formatters

### LINKED ISSUE #158: TodoEditor shows humanized datetimes in due field, may not be able to parse them back
The textfield is prefilled with something like "Today 23:00", but it's unclear (or at least not guaranteed) that this is going to be parsed again correctly.

~~We need a flag for `format_due` to disable human datetimes and use that within TodoEditor.~~

untitaker:
Fix #158

Doesn't do anything against #159, because that only crashes because
PorcelainFormatter has no `parse_datetime` method. Could we fix this by having
`PorcelainFormatter` subclass `TodoFormatter`? Whatever we do there, we should find a way to ensure `PorcelainFormatter` and `TodoFormatter` have the same methods. Sounds like a job for ABCs, but those have runtime overhead.

### DIFF
--- a/todoman/cli.py
+++ b/todoman/cli.py
@@ -171,7 +171,7 @@ def cli(click_ctx, color, porcelain):
     if len(paths) == 0:
         click.echo("No lists found matching {}, "
                    "create a directory for a new list"
-                   .format(config["main"]["path"]))
+                   .format(ctx.config["main"]["path"]))
         ctx.exit(1)
 
     ctx.db = Database(paths, ctx.config['main']['cache_path'])
@@ -285,8 +285,7 @@ def done(ctx, ids):
     Mark a task as done.
     '''
     for id in ids:
-        database = ctx.db
-        todo = database.todo(id)
+        todo = ctx.db.todo(id)
         todo.is_completed = True
         todo.save()
         click.echo(ctx.formatter.detailed(todo))
--- a/todoman/cli.py
+++ b/todoman/cli.py
@@ -1,6 +1,7 @@
 import functools
 import glob
 import locale
+import sys
 from datetime import timedelta
 from os.path import expanduser, isdir
 
@@ -9,7 +10,7 @@ import click_log
 
 from . import model
 from .configuration import ConfigurationException, load_config
-from .model import Database, FileTodo
+from .model import cached_property, Database, FileTodo
 from .ui import EditState, PorcelainFormatter, TodoEditor, TodoFormatter
 
 TODO_ID_MIN = 1
@@ -22,19 +23,20 @@ def _validate_lists_param(ctx, param=None, lists=None):
 
 
 def _validate_list_param(ctx, param=None, name=None):
+    ctx = ctx.find_object(AppContext)
     if name is None:
-        if 'default_list' in ctx.obj['config']['main']:
-            name = ctx.obj['config']['main']['default_list']
+        if 'default_list' in ctx.config['main']:
+            name = ctx.config['main']['default_list']
         else:
             raise click.BadParameter(
                 "{}. You must set 'default_list' or use -l."
                 .format(name)
             )
-    for l in ctx.obj['db'].lists():
+    for l in ctx.db.lists():
         if l.name == name:
             return l
     else:
-        list_names = [l.name for l in ctx.obj['db'].lists()]
+        list_names = [l.name for l in ctx.db.lists()]
         raise click.BadParameter(
             "{}. Available lists are: {}"
             .format(name, ', '.join(list_names))
@@ -42,20 +44,23 @@ def _validate_list_param(ctx, param=None, name=None):
 
 
 def _validate_date_param(ctx, param, val):
+    ctx = ctx.find_object(AppContext)
     try:
-        return ctx.obj['formatter'].parse_datetime(val)
+        return ctx.formatter.parse_datetime(val)
     except ValueError as e:
         raise click.BadParameter(e)
 
 
 def _validate_priority_param(ctx, param, val):
+    ctx = ctx.find_object(AppContext)
     try:
-        return ctx.obj['formatter'].parse_priority(val)
+        return ctx.formatter.parse_priority(val)
     except ValueError as e:
         raise click.BadParameter(e)
 
 
 def _validate_start_date_param(ctx, param, val):
+    ctx = ctx.find_object(AppContext)
     if val is None:
         return val
     if val.startswith('before '):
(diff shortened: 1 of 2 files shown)


todoman/cli.py
```diff
@@ -97,6 +102,35 @@ def command_wrap(*a, **kw):
...
 
 
+class AppContext:
+    def __init__(self):
+        self.config = None
+        self.db = None
+        self.porcelain = False
+
+    @cached_property
+    def ui_formatter(self):
```
hobarrera: I prefer the name `formatter` here.

todoman/cli.py
```diff
@@ -97,6 +102,35 @@ def command_wrap(*a, **kw):
     return command_wrap
 
 
+class AppContext:
+    def __init__(self):
+        self.config = None
+        self.db = None
+        self.porcelain = False
+
+    @cached_property
```
hobarrera: Are there any advantages to lazy-initialization? Is there any command that **doesn't** use a formatter?

todoman/cli.py
```diff
@@ -115,33 +149,23 @@ def command_wrap(*a, **kw):
...
-    color = color or config['main']['color']
+    color = color or ctx.config['main']['color']
     if color == 'always':
-        ctx.color = True
+        click_ctx.color = True
     elif color == 'never':
-        ctx.color = False
+        click_ctx.color = False
+
+    ctx.porcelain = porcelain
```
hobarrera: I'd much rather pass the class and abstract knowledge of porcelainness from the context object:

```python
ctx.formatter_class = PorcelainFormatter
```

Then instantiate the formatter as-needed.  All formatter should have the same signature anyway (including constructor).

todoman/cli.py
```diff
@@ -115,33 +149,23 @@ def command_wrap(*a, **kw):
               'remain stable regardless of configuration or version.')
 @click.pass_context
 @click.version_option(prog_name='todoman')
-def cli(ctx, color, porcelain):
+def cli(click_ctx, color, porcelain):
+    ctx = click_ctx.ensure_object(AppContext)
```
hobarrera: I don't understand what `ensure_object` does, can you explain it a bit?

todoman/cli.py
```diff
@@ -224,49 +247,49 @@ def edit(ctx, id, todo_properties, interactive):
...
         todo.save()
         if old_list.name != todo.list.name:
-            database.move(todo, todo.list, from_list=old_list)
-        click.echo(ctx.obj['formatter'].detailed(todo))
+            ctx.db.move(todo, todo.list, from_list=old_list)
+        click.echo(ctx.formatter.detailed(todo))
     else:
         click.echo('No changes.')
-        ctx.exit(1)
+        sys.exit(1)
```
hobarrera: So, `ctx.exit` no longer works? Does this affect us at all?

todoman/cli.py
```diff
@@ -224,49 +247,49 @@ def edit(ctx, id, todo_properties, interactive):
...
+@pass_ctx
 @click.argument('ids', nargs=-1, required=True, type=click.IntRange(0))
 def done(ctx, ids):
     '''
     Mark a task as done.
     '''
     for id in ids:
-        database = ctx.obj['db']
+        database = ctx.db
         todo = database.todo(id)
```
hobarrera: Just use `ctx.db` here and get rid of the previous line.

todoman/cli.py
untitaker: What would you name the current `formatter` property then?

todoman/cli.py
untitaker: There are few commands that use anything other than the default formatter.

todoman/cli.py
untitaker: It goes through the `ctx.obj` of all contexts and parent contexts, and sets the current `ctx.obj` to `AppContext()` if nothing is found.

todoman/cli.py
untitaker: >All formatter should have the same signature anyway (including constructor).

That currently isn't the case, and I don't see a good reason for this to change.

todoman/cli.py
untitaker: `ctx` is now our `AppContext`, not `click.Context`. I didn't find it very apealing to pass the click ctx just because of that.

todoman/cli.py
untitaker: Done

todoman/cli.py
hobarrera: Sorry, the `formatter` attribute below slipped completely past me. I blame the lack of sleep.

todoman/cli.py
hobarrera: Sorry, the `formatter` attribute below slipped completely past me. I blame the lack of sleep.

todoman/cli.py
hobarrera: :+1:

todoman/cli.py
hobarrera: I understood this bit, I just wanted to make sure this had no downsides.

todoman/cli.py
```diff
@@ -224,49 +247,48 @@ def edit(ctx, id, todo_properties, interactive):
             setattr(todo, key, value)
 
     if interactive or (not changes and interactive is None):
-        ui = TodoEditor(todo, ctx.obj['db'].lists(), ctx.obj['formatter'])
+        ui = TodoEditor(todo, ctx.db.lists(), ctx.ui_formatter)
```
hobarrera: TBH, I'm not very convinced by this. I'd much rather disallow `interactive` and `porcelain` together (since they don't make too much sense anyway).

todoman/cli.py
hobarrera: Having the same API on both assures that we can interchange them without anything breaking.

untitaker: There's a common interface both classes implement, doesn't mean that all methods have to have the same signature.

You gain something from having the same instancemethods on both classes because that's what TodoEditor etc call, but you don't gain anything from doing this at init time. The code doesn't get significantly shorter.

todoman/cli.py
untitaker: You can use `todo edit` *partially* interactively as well, e.g. use `--due` to override that field, but also use `-i` to launch the editor afterwards (with the new values)

hobarrera: Does this make sense? Do you think people will do that? Do we *care* about that scenario?

untitaker: I think the mix between interactive and non-interactive *only* makes sense using `--porcelain`. Usecase is a script that launches `todo new` with certain default values. `todo edit` works the same way, but with more default values.

hobarrera: Yup, quite right. Other apps, or scripts could really really use this. :+1:

PR #161, opened 2017-03-02
Implement --humanize

### LINKED ISSUE #128: Humanize all date times?
I'd like to evaluate humanizing all date times (probably with [humanize](https://github.com/jmoiron/humanize)).

* Not the default (due to loss of precision).
* Probably some `-h/--humanize` flag.

We'd have to look at pitfalls, and potential issues that may be a bit non-obvious.

### LINKED ISSUE #159: `todo --porcelain edit ...` crashes

hobarrera:
* [x] Tests
* [ ] Tests
* [ ] Tests

Fixes #159 
Closes #128

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,6 +2,7 @@ atomicwrites
 click>=6.0
 click_log>=0.1.3
 configobj
+humanize
 icalendar
 parsedatetime
 python-dateutil
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -9,6 +9,7 @@ from dateutil.tz import tzlocal
 from hypothesis import HealthCheck, settings, Verbosity
 
 from todoman import model
+from todoman.ui import TodoFormatter
 
 
 @pytest.fixture
@@ -88,6 +89,11 @@ def todo_factory(default_database):
         return todo
 
     return inner
+
+
+@pytest.fixture
+def default_formatter():
+    return TodoFormatter("%Y-%m-%d", "%H:%M", ' ')
 
 
 settings.register_profile("ci", settings(
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -9,7 +9,7 @@ from dateutil.tz import tzlocal
 from hypothesis import HealthCheck, settings, Verbosity
 
 from todoman import model
-from todoman.ui import TodoFormatter
+from todoman.ui import DefaultFormatter
 
 
 @pytest.fixture
@@ -93,7 +93,7 @@ def todo_factory(default_database):
 
 @pytest.fixture
 def default_formatter():
-    return TodoFormatter("%Y-%m-%d", "%H:%M", ' ')
+    return DefaultFormatter("%Y-%m-%d", "%H:%M", ' ')
 
 
 settings.register_profile("ci", settings(
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -491,4 +491,12 @@ def test_sort_mixed_timezones(runner, create):
     assert 'second' in result.output.splitlines()[0]
     assert 'first' in result.output.splitlines()[1]
 
+
+def test_humanize_interactive(runner):
+    result = runner.invoke(cli, ['--humanize', '--porcelain', 'list'])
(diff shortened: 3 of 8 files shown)


untitaker:
`TodoFormatter` already has code for humanizing datetimes. What about that?

hobarrera:
The main difference is that the existing code merely replaces a date with "tomorrow", making it more readable but with no loss in precision. The `HumanizedFormatter` has output with less precision, and exact dates/times are lost.

untitaker:
My point is that it's not very intuitive for humanized datetimes to be shown when `--humanized` is not used.

hobarrera:
Yeah, I agree with that. Maybe the docs could cover that a bit? Do you have any other suggestion?

A second choice is to rename `--humanize` to `--fuzzy`.

untitaker:
Why not just remove the humanizing code from the default formatter?

hobarrera:
> Why not just remove the humanizing code from the default formatter?

Don't really see why. :(

untitaker:
I mean that only when `--humanize` is used, the date is replaced with `Today` and so on. I.e. remove the special datetimes dict and all that.

hobarrera:
Nope, I don't think I understand.

untitaker:
With this PR we have two ways to humanize datetimes, once using `humanize`, once using #108. I think there's no major difference between being precise or not. The bigger difference is between formatting with the normal `date_format` from the config... and not doing that.

hobarrera:
I'm not sure if you dislike the result or the implementation (or both).

untitaker:
The intended UI

hobarrera:
So, would you rather not show "Tomorrow/"Today" in non-humanized form (for the basic formatter)? Or do you dislike this PR entirely?

What would you change for it?

untitaker:
When `humanize` setting is false, don't use the strings "Today" or "Tomorrow"
at all, i.e. use pre-#108 behavior.

Otherwise use the humanize package.

hobarrera:
> When `humanize` setting is false, don't use the strings "Today" or "Tomorrow" at all, i.e. use pre-#108 behavior.

Why do you feel that's a bad idea? (I'm honestly just trying to understand your POV).

untitaker:
I'm not sure how to express it better: It feels bad to have two implementations
for humanizing datetimes, first #108 and now using the `humanize` pkg. Two
implementations that basically serve the same purpose (again I don't consider
the distinction "fuzzy" vs "exact" important). IMO one of them has to go.

hobarrera:
PR updated (and a lot has been rewritten).

untitaker:
This reintroduces the bug #160 fixed, because now we might end up using humanized datetimes inside TodoEditor again.

hobarrera:
Oh, you're right; I just need to modify `formatter` and never `ui_formatter`, right?

untitaker:
Right, but we could also stick closer to the original design and have only two `TodoFormatters`, with formatting methods having a `humanize` flag.

hobarrera:
> Right, but we could also stick closer to the original design and have only two TodoFormatters, with formatting methods having a humanize flag.

I think this is actually a perfect scenario to have to classes because we have perfect polymorphism: both implement the same API, but have different implementations.

untitaker:
In either case you're now mixing two approaches: You're passing `porcelain` as a flag instead of `formatter_class=PorcelainFormatter`, at the same time you don't pass `humanize` but influence behavior using `formatter_class`.

hobarrera:
Yup, that's ugly. Updated.

hobarrera:
I prefer `DefaultFormatter` because it's also the one used by default and that makes it a wee bit more obvious.

Regarding the property names: I think they're quite representative, but I'm open to better suggestions.

hobarrera:
Rebased. I actually found another bug we had hiding there (which was made evident due to the merge), so I fixed it and added a few tests.

PR #166, opened 2017-03-03
Add support for editing and printing categories

Sakshisaraswat:
Add support for categories and written some test cases for it


todoman/cli.py
```diff
@@ -171,10 +171,12 @@ def cli(ctx, color, porcelain):
 @click.argument('summary', nargs=-1)
 @click.option('--list', '-l', callback=_validate_list_param,
               help='The list to create the task in.')
+@click.option('--category',
+              help='Categories of todo')
```
hobarrera: Please change this to `A comma-separated list of categories`, since the comma-separated bit won't be obvious to users (though it definitely is to you and me 😄  ).

todoman/model.py
```diff
@@ -165,6 +165,10 @@ def categories(self):
     def raw_categories(self):
         return self.todo.get('categories', '')
 
+    @raw_categories.setter
+    def raw_categories(self, categories):
```
hobarrera: :+1:

todoman/cli.py
hobarrera: Please have a quick look at `_todo_property_options`. We basically have `option`s which are common to `new` and `edit` there, so they apply to both places. This `option` should be moved there to keep the code simpler.

todoman/ui.py
```diff
@@ -66,6 +66,10 @@ def __init__(self, todo, lists, formatter):
             parent=self,
             edit_text=todo.location
         )
+        self._categories = widgets.ExtendedEdit(
+            parent=self,
+            edit_text=','.join(todo.categories)
```
hobarrera: Why didn't you use `raw_categories` here?

untitaker:
I resolved some conflicts that resulted from #160, you may have to pull before you can push.

todoman/ui.py
```diff
@@ -313,6 +325,14 @@ def format_datetime(self, dt, humanize=True):
             self._format_time(time_part)
         )))
 
+    def parse_category(self, categories):
+        if not categories:
+            return None
+        if ',' in categories:
```
untitaker: You can just do `return categories.split(',')` here:

    >>> "foo".split(',')
    ["foo"]

untitaker:
@Sakshisaraswat I don't understand what you meant in IRC, the tests pass.

Sakshisaraswat:
codecov/patch 
this is not passed

untitaker:
Ah, it appears that you have no testcase for [this line of code](https://github.com/pimutils/todoman/pull/166/files#diff-6ad87ea802c8a8dc00c82d7eea9889d5R87), i.e. if you pass an invalid category.

CHANGELOG.rst~
```diff
@@ -0,0 +1,72 @@
+Changelog
```
untitaker: You accidentally committed this file, please remove it.

todoman/ui.py
```diff
@@ -55,6 +55,10 @@ def __init__(self, todo, lists, formatter):
             parent=self,
             edit_text=todo.location
         )
+        self._categories = widgets.ExtendedEdit(
+            parent=self,
+            edit_text=','.join(todo.raw_categories)
```
untitaker: Since `todo.raw_categories` is already a string that contains comma-separated categories, `','.join(..)` is not necessary.

todoman/cli.py
```diff
@@ -80,6 +80,14 @@ def _validate_start_date_param(ctx, param, val):
         raise click.BadParameter(e)
 
 
+def _validate_category(ctx, param, val):
+    ctx = ctx.find_object(AppContext)
+    try:
+        return ctx.formatter.parse_category(val)
+    except ValueError as e:
```
hobarrera: I don't think we can ever raise this exception in `Formatter.parse_category`.

untitaker: For consistency I'd vote for leaving it in though. When editing `parse_category`, I would expect that `ValueError` is handled like it always is.

hobarrera: It's dead/unreachable code, handling an error for an impossible scenario.

untitaker: Again, I'm pretty sure we will introduce a UI bug when editing `parse_category` in the future.

hobarrera: I don't see how this helps catch that very-specific bug. If a `ValueError` is raised, or a `BadParameter` is raised, we're still not catching that possible bug, because no tests will cover that new scenario.

untitaker: We don't have to catch a bug because none ever occurs...

hobarrera: TBH, I see no point in catching is error. It's just here for posterity's sake _in case_ we make an error in the code elsewhere, but will only catch one error and show another (which won't help either). It's also code that cannot be tested, until we break something else (and even then, we have to manually find that other breakage ourselves, because no tests will).

untitaker: Ok, I don't care.

hobarrera:
I've resolved conflicts for this PR, since I pushed changes to `master` that resulted in conflict for you.

hobarrera:
Closing due to inactivity. Feel free to reopen if you're still interested on working on this.

PR #170, opened 2017-03-04
Properly parse empty priority in the UI

### LINKED ISSUE #156: TodoEditor requires setting a non-empty priority when saving
It should default to `none`

hobarrera:
Fixes #156

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -69,6 +69,19 @@ def now_for_tz():
     return inner
 
 
+@pytest.fixture
+def todo_factory(default_database):
+    def inner():
+        todo = model.FileTodo()
+        todo.list = list(default_database.lists())[0]
+        todo.summary = 'YARR!'
+        todo.save()
+
+        return todo
+
+    return inner
+
+
 settings.register_profile("ci", settings(
     max_examples=1000,
     verbosity=Verbosity.verbose,
--- a/tests/test_ui.py
+++ b/tests/test_ui.py
@@ -1,26 +1,24 @@
-from todoman.model import FileTodo
-from todoman.ui import PorcelainFormatter, TodoEditor
+import pytest
+from urwid import ExitMainLoop
+
+from todoman.ui import TodoEditor, TodoFormatter
+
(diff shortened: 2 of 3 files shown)


hobarrera:
This contains the first *real* UI test. :tada:

hobarrera:
Moving forward, I think it makes sense to have a `test_formatter` fixture which just returns `TodoFormatter` with the `DATE_FORMAT` and `TIME_FORMAT` we're using for tests.

hobarrera:
@untitaker I hope you don't take reviewing this as an imposition, I just feel you'd like to review this (mostly because it sets a new example for tests) and you might have some comments or suggestions (or not).

tests/conftest.py
```diff
@@ -69,6 +69,16 @@ def inner(tz='CET'):
     return inner
 
 
+@pytest.fixture
+def blank_todo(default_database):
+    todo = model.FileTodo(new=True)
```
untitaker: `FileTodo`s are `new` by default.

tests/conftest.py
```diff
@@ -69,6 +69,16 @@ def inner(tz='CET'):
     return inner
 
 
+@pytest.fixture
+def blank_todo(default_database):
```
untitaker: I think it's more generally usable if this fixture returns a function that creates todos (eventually with a parameter for specifying list)

hobarrera:
> Why make a difference between None and 0? They have the same semantics.

Basically one is "don't store anything in the file", and the other "store 0".  
I think this is a bit important, because if we edit a file, I'd like to avoid adding `PRIORITY:0\n` to it if the user didn't touch that field.

hobarrera:
> This reintroduces the bug fixed by #160 because now we might again end up using humanized datetimes when prefilling TodoEditor.

Are you sure this comment was meant for this PR?

untitaker:
>Are you sure this comment was meant for this PR?

That was a mistake I think

untitaker:
It's up to you whether you want to make the model more complex for a nicer file diff.

PR #171, opened 2017-03-04
Simplify argument passing from list() to todo()

hobarrera:


hobarrera:
Ugh, this is old as already merged, guh.

PR #180, opened 2017-03-07
Warn when a file contains multiple VTODO entries

hobarrera:
Also, don't load these at all, since read-writing them is destructive (maybe we could deal with this by setting a READONLY flag?).

Also debug a bit more.

Pending:
* [ ] Tests!
* [ ] Cached files don't WARN again. :(

### DIFF
--- a/todoman/model.py
+++ b/todoman/model.py
@@ -306,17 +306,23 @@ class FileTodo(Todo):
         with open(path, 'rb') as f:
             cal = f.read()
             cal = icalendar.Calendar.from_ical(cal)
-            try:
-                component = cal.walk('VTODO')[0]
-                todo = cls(
-                    new=False,
-                    todo=component,
-                    filename=os.path.basename(path),
-                )
-                todo.id = id
-                return todo
-            except IndexError:
-                pass
+            components = cal.walk('VTODO')
+            if not components:
+                logger.debug('No VTODO in %s', path)
+                return
+
+            if len(components) > 1:
+                logger.warn('Multiple VTODO entries in %s!', path)
+                return
+
+            todo = cls(
+                new=False,
+                todo=components[0],
+                filename=os.path.basename(path),
+            )
+            todo.id = id
+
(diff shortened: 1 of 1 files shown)


hobarrera:
Okay, so this isn't any good because we don't warn after the first time due to cache. Maybe we have to cache the error (ugh), or just mark it read-only (that still won't warn again)

 :(

untitaker:
If you don't care about displaying them, you could just skip caching them every time.

hobarrera:
> If you don't care about displaying them, you could just skip caching them every time.

That's what I'd like, but by the time we're reading files they've already been cached. :(

untitaker:
Return a list of `Todo`s from `FileTodo.from_file` and put the warning into `Database.update_cache`?

hobarrera:
> Return a list of `Todo`s from `FileTodo.from_file` and put the warning into `Database.update_cache`?

I like that idea (especially because we can read *all* todos), even multiple per file (even though this violates vdir!).

Basically:
* `from_file` return multiples `FileTodo`s.
* Add a `read_only` attribute and cache column (default false).
* Add a warnings table with todo warnings ("notes"?).
* Print warnings whenever we show that individual todo.
* Show an icon in lists to identify read-only todos.
* Disallow editing of these todos (error out).

A command to split them into separate files might also be helpful.

Does that sound okay?

untitaker:
I rather thought about just printing out the warning and skipping caching them so the warning shows every time.

hobarrera:
> I rather thought about just printing out the warning and skipping caching them so the warning shows every time.

Yeah, a like the idea, but I'm not sure how to implement it: once we're at this point we've already cached the file (but not the todo). This would make todoman skip it next time (much like file with no `VTODO` entries).

To be clear: I like the idea, I'm not sure how to implement it cleanly.

untitaker:
See https://github.com/pimutils/todoman/pull/188.

PR #188, opened 2017-03-09
Show error about files with not exactly one todo

untitaker:


todoman/model.py
```diff
@@ -306,17 +306,14 @@ def from_file(cls, path, id=None):
         with open(path, 'rb') as f:
             cal = f.read()
             cal = icalendar.Calendar.from_ical(cal)
-            try:
-                component = cal.walk('VTODO')[0]
+            for component in cal.walk('VTODO'):
```
hobarrera: I'd rather `raise` in this method, a bit like  #180.

Also, I'd rather not support more than one-per-file by default, since this is vdir-incompliant (and vdirsyncer-incompatible).

untitaker:
I refactored everything. Warnings about todos are now shown everytime, at the same time they're accessible as read-only. An exception is thrown when the user tries to edit. However, the exception handling could be prettier.

untitaker:
Fixed the error UX

todoman/model.py
```diff
@@ -702,7 +716,20 @@ def expire_lists(self, paths):
     def delete_list(self, name):
         self._conn.execute("DELETE FROM lists WHERE lists.name = ?", (name,))
 
-    def todo(self, id):
+    def todo(self, id, from_db=False):
```
hobarrera: I'd rather call this flag `read_only`, because rather than expose *how* we read the `todo`, that naming reflects what the caller cares about (_"I want to edit this todo"_).

Or did I miss some scenarios where this wouldn't make sense?

What do you think?

todoman/model.py
```diff
@@ -18,6 +18,10 @@ class NoSuchTodo(Exception):
     pass
 
 
+class ReadOnlyTodo(Exception):
```
hobarrera: 👌🏻

todoman/cli.py
```diff
@@ -282,21 +307,19 @@ def edit(ctx, id, todo_properties, interactive):
 @cli.command()
 @pass_ctx
 @with_id_arg
+@catch_errors
```
untitaker: I don't know how to do this. I think it would require a subclass of `click.Group

todoman/model.py
untitaker: Agreed

todoman/cli.py
hobarrera: Yeah, leave it as-is for now, I've some ideas for this and `pass_ctx`, but they're out of scope for this.

untitaker:
No, though it would probably better if we did. I'll come up with something in the next days

On 10 March 2017 15:12:20 CET, Hugo Osvaldo Barrera <notifications@github.com> wrote:
>hobarrera approved this pull request.
>
>If I understood correctly we *do not* warn on `show`, right? (I've no
>issue with this, just wanting to make sure).
>
>LGTM
>
>
>
>-- 
>You are receiving this because you authored the thread.
>https://github.com/pimutils/todoman/pull/188#pullrequestreview-26283922

-- 

untitaker:
Nevermind, I can't find an efficient way to do this. Let's merge this.

hobarrera:
My only idea was to have a `warnings` table, and see if a read todo is present there when reading. Not sure how efficient that is (and it's probably ugly too).

Feel free to merge this. 🚀

PR #192, opened 2017-03-18
Simplify the Todo model

### LINKED ISSUE #190: Remove separation of `Todo` and `FileTodo`
Users should be able of completely ignore if a `Todo` comes from the cache or not mostly. This might open the doors for alternate backends later on too (though it's not a priority).

### LINKED ISSUE #191: Remove the `icalendar` instance backing `Todo`s
As discussed on IRC, remove the `icalendar` instance backing (eg: `VTODO`) backing every `Todo`. This'll simplify some code, and provide a clearer separation of read/write.

hobarrera:
Simplify the todo model, and much of the read/write dancing we do. Also include a summary of how this flow is in the docuemntation.

There are *extremely* few changes outside `models.py`, to evidence how this is mostly an internal refactor with little consequence. Only a not-very-good test was altered.

Fixes #190 
Fixes #191

### DIFF
--- a/todoman/model.py
+++ b/todoman/model.py
@@ -749,25 +749,29 @@ class Cache:
 class List:
 
     def __init__(self, name, path, colour=None):
-        self.name = name
         self.path = path
-        self.colour = colour
+        self.name = name or List.name_for_path(path)
+        self.colour = colour or List.colour_for_path(self.path)
 
-    @cached_property
-    def color_raw(self):
-        '''
-        The color is a file whose content is of the format `#RRGGBB`.
-        '''
-
+    @staticmethod
+    def colour_for_path(path):
         try:
-            with open(os.path.join(self.path, 'color')) as f:
+            with open(os.path.join(path, 'color')) as f:
                 return f.read().strip()
         except (OSError, IOError):
-            logger.debug('No colour for list %', self.path)
+            logger.debug('No colour for list %s', path)
+
+    @staticmethod
+    def name_for_path(path):
+        try:
+            with open(os.path.join(path, 'displayname')) as f:
+                return f.read().strip()
+        except (OSError, IOError):
+            return split(normpath(path))[1]
 
     @cached_property
     def color_rgb(self):
-        color = self.colour or self.color_raw
+        color = self.colour
         if not color or not color.startswith('#'):
             return
 
@@ -810,9 +814,9 @@ class Database:
 
         for path in self.paths:
             list_name = self.cache.add_list(
-                self._list_name(path),
+                List.name_for_path(path),
                 path,
-                self._list_colour(path),
+                List.colour_for_path(path),
             )
             for entry in os.listdir(path):
                 if not entry.endswith('.ics'):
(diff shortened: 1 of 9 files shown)


todoman/model.py
```diff
@@ -87,263 +92,245 @@ def __init__(self, todo=None, filename=None, mtime=None, new=False):
...
+        'percent_complete': 0,
+        'priority': 0,
+        'sequence': 0,
+        'status': 'NEEDS-ACTION',
+        'summary': '',
+        'uid': None,
+        'id': None,  # Not the right place
+    }
+
+    KNOWN_FIELDS = DEFAULT_VALUES.keys()
```
untitaker: Why is this a separate value. I understand that it makes the code prettier, but you might as well rename `DEFAULT_VALUES` to `KNOWN_FIELDS`.

todoman/model.py
```diff
@@ -87,263 +92,245 @@ def __init__(self, todo=None, filename=None, mtime=None, new=False):
...
+    def __getattr__(self, name):
+        if name in Todo.KNOWN_FIELDS:
+            if name in self._attributes:
+                return self._attributes[name]
+            else:
+                return Todo.DEFAULT_VALUES[name]
+        raise AttributeError("'{}' has no attribute '{}'".format(self, name))
+
+    def __setattr__(self, name, value):
+        if name in Todo.KNOWN_FIELDS:
```
untitaker: I imagined this code to be more like `TodoEditor`: Simple attributes on the `Todo` instance.

todoman/model.py
hobarrera: TBH, probably a result of intermediate refactors. I'll get rid of this field altogether. :+1:

todoman/model.py
hobarrera: Yeah, the initial version of this code didn't allow for that, but, as it it right now, that makes more sense.  This whole bit looks over-engineered now that you mention this.

hobarrera:
Thanks for the feedback, @untitaker, I know you'd have additional improvements for this! 😉

todoman/model.py
hobarrera: Trying to do this actually unveiled how fragile all the dealing with defaults is (eg: priority 0 vs None, etc).

I'll put some work into this, since I don't like this fragility.

hobarrera:
I've pushed 988c94c to make `Todo` attributes plain attributes, but I'm not 100% convinced. The big downside is that there's no set-time normalization, so usages because a little trickier. For example, setting `priority = 0` and `priority = None` might have _slightly_ different effects in some corner cases.

A concrete example is that setting `description = None` makes the editor crash. The previous behaviour skipped this assignment and retained the former value, `''`.

untitaker:
In that case, make them properties with separate logic each. Very verbose code but easier to understand.

On 18 March 2017 23:16:04 CET, Hugo Osvaldo Barrera <notifications@github.com> wrote:
>I've pushed 988c94c to make `Todo` attributes plain attributes, but I'm
>not 100% convinced. The big downside is that there's no set-time
>normalization, so usages because a little trickier. For example,
>setting `priority = 0` and `priority = None` might have _slightly_
>different effects in some corner cases.
>
>A concrete example is that setting `description = None` makes the
>editor crash. The previous behaviour skipped this assignment and
>retained the former value, `''`.
>
>-- 
>https://github.com/pimutils/todoman/pull/192#issuecomment-287578806

-- 

hobarrera:
@untitaker I just pushed 1016b75, which is an intermediate compromise (inspired by your previous comments, BTH).

Rather than have a dozen setters (plus the setters which we don't even need!), just override setting in one place, and group attributes by type. 😄 

This is a lot more resilient than the both the original and previous version, and I feel more confident it it's type-management.

hobarrera:
Given the way the code is laid out, I think we can (and also could before) write to a multi-VTODO files (given that we only replace a single component inside the file). After we merge this, I'll look into adding tests for that (but will leave a warning to discourage that!).

PR #193, opened 2017-03-18
Don't require quoting args of --start

hobarrera:
By design, we used to have require that the arguments of --start be quoted, eg:

    todo list --start "after 2017-03-18"

Drop this requirement, and treat both things as two different arguments.


untitaker:
Now the old syntax doesn't work. I'm not sure I like this.

hobarrera:
The old syntax was never released. And I think it's kinda counter-intuitive that *this* particular parameter requires quoting/escaping.

untitaker:
I don't think there's any other option that allows this kind of behavior. The situation is different for the argument of `todo new`.

hobarrera:
No other option (command) takes two arguments either. I think that, considering it takes two args, the treat the as such. Not only because of the code-cleaness, but also because I think it's what's most intuitive.

Just for the record, I though the option was totally broken when testing it, and had to look at the tests to realize I needed to escape the args.

Issue #184, opened 2017-03-08
Review due/start/alarms


hobarrera:
I'd like to review how we use due/start/alarms.

Some ideas are:

* Have a filter to filter by "alarms soon", so that we can have some reminder app poll that.
* Support for adding/removing alarms (also, some default that matches the due date?)
* Filter tasks by `dtstart`. Personally, I'd like to, by default, filter out tasks that have a `start` date in future (this will probably have to be configurable).

As far as UI goes, the I'd add a middle column with alarm and recurrances.

This requires a lot of consideration, and discussion. I'll leave this open for a while for suggestions and criticism.

Issue #17, opened 2016-04-17
Done seems to delete repeating reminders


hobarrera:
This issue was moved from the issue tracker at GitLab.

---

When I use todoman to mark repeating reminders "done", the repeating reminders disappear entirely, and are not rescheduled.

I am using todoman version 1.5.2.

hobarrera:
This was also reported on a different issue by @languitar

rajalakshmi-v15:
hi,
I want to help fix this issue. I am new to this project and am interested in working on this particular issue. Please guide me.

untitaker:
I'm not entirely sure what is needed for this issue, but I consider it to be very long-term (and very hard). Not sure if having the rgsoc label on it is a good idea.

rajalakshmi-v15:
We hoped to work on this issue for RGSoc. Could you suggest any other issue which might be suitable to be worked on during the summer?

untitaker:
We currently have more participants than issues, though I think #62 isn't taken. But I'd wait for @hobarrera's response.

hobarrera:
This specific issue is a bit long, so it might be suitable to work during the summer (as a long-term one), *however* I do suggest that you pick one labelled `lv:easy` first to get acquainted with todoman.

Also, @untitaker is right, we also have to define a few thing before we can tackle this one.

hobarrera:
So, currently, for this task, we need to:

* Design how we'll model recurring tasks.
* Investigate how we mark done instances (I'd have to re-read the RFC, but, so far, I didnt' find anything useful) I might opt for using some `X-Last-Done`, which keeps the datetime of the last instance marked as done.

I believe that iOS deals with this by using alarms, which we also don't support, but should

hobarrera:
So what iOS does is pretty simple (I tried this with a monthly recurring todo I actually needed). It simply creates a normal recurring event, and when I mark it done here's what happens:

* A copy of the vtodo is made, but sets the copy's DTSTART and DUE to the next instance's date.
* It remove the RRULE from the original copy.
* The copy gets a different UID.
* It sets the original copy's COMPLETED and PERCENT-COMPLETE field, as well as updating the STATUS field.
* The ALARM for the new copy also has it's date pushed to the next instance, but the AKNOWLEDGED attribute is set for it (IMHO, the old copy should have this set).

What I don't understand is how iOS recognized both as the same recurring todo. It probably doesn't, since clicking _show completed_ shows me the done copy, while I can see the incomplete copy above.

[...]

PR #200, opened 2017-03-21
Add support for the `dtstart` field

hobarrera:
Basically, according to the spec:

* `dtstart` defines when a todo should/can be started.  
  If it's `NULL`, it can be considered current every day.
* `due` defined when a todo is overdue.

To comply with this, hide todos which have a `dtstart` in the future by default (this is configurable).

1. I'm very much willing to listen to better names for the setting/flag. `today` is the best I could think of and sucks. `started` is a bit ambiguous, because started todos are those with `percent_complete > 0`. `workable` might to but is also really unintuitive.
1. I really prefer to keep this default, but if there's strong opposition (@untitaker?), I can revert it.
1. I'm inclined to remove `--today` since it looks superfluous (unless we revert the default).

This allows on to create far-into-the-future todos, and not have them clutter up the default view. It also makes recurring todos less painful.

Somewhat related to #184. Paves the way for #17.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,8 @@ releases, in reverse chronological order.
 v3.0.0
 ------
 
+* Added a ``today`` setting and flag to show only todos which can be started
+  today (eg: don't start in the future).
 * Add the ``--humanize`` to show friendlier date times (eg: ``in 3 hours``).
 * Dropped ``--urgent`` and introduced ``--priority``
 * Added support for filtering by priority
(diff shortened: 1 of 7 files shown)


untitaker:
I wouldn't hide todos that start in the future, but only put dtstart into the
default sort order.

hobarrera:
I just realized that the default I'd left was actually what you prefer. 😆 
I've added tests to make sure we don't inadvertently change that.

PR #208, opened 2017-03-23
Windows CI?

hobarrera:
This will likely fail and I'll give up in the end (most likely), but let's give it a try anyway.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -6,14 +6,6 @@ environment:
     CMD_IN_ENV: "cmd /E:ON /V:ON /C .\\.ci\\run_with_env.cmd"
 
   matrix:
-    - PYTHON: "C:\\Python27"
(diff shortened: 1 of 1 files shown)


hobarrera:
I'll be cancelling travis runs for this so as not to waste.

hobarrera:
I don't understand these errors, and a lot of them aren't even close to making sense. Other are related to line-endings and can be avoided by using `.strip()`:

```
>       assert cli_result.output == main_output.decode()
E       AssertionError: assert 'todoman, ver...2+ng973d06e\n' == 'todoman, vers...ng973d06e\r\n'
E         - todoman, version 2.1.0.post122+ng973d06e
E         + todoman, version 2.1.0.post122+ng973d06e
E         ?                                         +
tests\test_main.py:20: AssertionError
```

PR #213, opened 2017-03-24
OS X

hobarrera:
Two **tests** are failing, which might we worth looking into. CI is actually running :tada:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ python:
 
 env:
   matrix:
-    - BUILD=py
+    - TOXENV=py
   global:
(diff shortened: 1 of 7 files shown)


untitaker:
I suspect low mtime-precision, like in khal.

hobarrera:
**Thank you**, I'd never thought of that! I'll add some mechanism to for cache invalidation for this sort of thing and add it to these tests (basically, force a reload).

untitaker:
What about copying the `min_sleep_time` fixture from khal?

hobarrera:
That fixture makes sense to test natural cache expiration, but I don't want to include that everywhere. For most tests, cache invalidation and sleeping is a bit out of scope and I want to force a reload.

I **will** add new tests to make sure we expire the cache properly, and all that mtime fuzz.

hobarrera:
Maybe what I **really** want is a `load_todos` fixture, instead of calling `default_database.update_cache(); list(default_database.todos())` in half of the tests.

hobarrera:
:tada:

hobarrera:
These failure must be a fluke; that diff only changes the README.

PR #242, opened 2017-04-26
Choose priority when adding a new task from cli

### LINKED ISSUE #239: Choose priority when adding a new task from cli
Describtion sais it all, sth like:

```
    todo new --priority high 'Choose priority when adding a new task'
```

AnubhaAgrawal:
Fixes #239 Choose priority when adding a new task from cli

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -870,20 +870,8 @@ def test_invoke_invalid_command(runner, tmpdir):
     )
 
 
-def test_show_priority(tmpdir, runner, create):
-    create(
-        'test.ics',
-        'SUMMARY:harhar\n'
-        'Priority:high\n'
-    )
+def test_show_priority(runner, todo_factory, todos):
+    todo = todo_factory(summary='harhar\n', priority=1)
 
     result = runner.invoke(cli, ['show', '1'])
-    assert 'high' in result.output
-
-
-def test_priority(runner):
-    result = runner.invoke(cli, [
-        'new', '-l', 'default', '--priority', 'high', 'Priority Test'
-    ])
-    assert not result.exception
-    assert 'high' in result.output
+    assert '1' in result.output
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -874,16 +874,16 @@ def test_show_priority(tmpdir, runner, create):
     create(
         'test.ics',
         'SUMMARY:harhar\n'
-        'Priority:High\n'
+        'Priority:high\n'
     )
 
     result = runner.invoke(cli, ['show', '1'])
-    assert 'High' in result.output
+    assert 'high' in result.output
 
 
 def test_priority(runner):
     result = runner.invoke(cli, [
-        'new', '-l', 'default', '--priority', 'Chembur', 'Event Test'
+        'new', '-l', 'default', '--priority', 'high', 'Priority Test'
     ])
     assert not result.exception
-    assert 'Chembur' in result.output
+    assert 'high' in result.output
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -871,7 +871,7 @@ def test_invoke_invalid_command(runner, tmpdir):
 
 
 def test_show_priority(runner, todo_factory, todos):
-    todo = todo_factory(summary='harhar\n', priority=1)
+    todo_factory(summary='harhar\n', priority=1)
 
     result = runner.invoke(cli, ['show', '1'])
     assert '!!!' in result.output
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -875,3 +875,12 @@ def test_show_priority(runner, todo_factory, todos):
 
     result = runner.invoke(cli, ['show', '1'])
     assert '1' in result.output
+
+
+def test_priority(runner):
+    result = runner.invoke(cli, [
+        'new', '-l', 'default', '--priority', 'high', 'Priority Test'
+    ])
+
+    assert 'high' not in result.output
+
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -868,3 +868,22 @@ def test_invoke_invalid_command(runner, tmpdir):
     assert (
         'Error: Invalid setting for [main][default_command]' in result.output
     )
+
+
+def test_show_priority(tmpdir, runner, create):
+    create(
+        'test.ics',
+        'SUMMARY:harhar\n'
+        'Priority:High\n'
+    )
+
+    result = runner.invoke(cli, ['show', '1'])
+    assert 'High' in result.output
+
+
+def test_priority(runner):
+    result = runner.invoke(cli, [
+        'new', '-l', 'default', '--priority', 'Chembur', 'Event Test'
+    ])
+    assert not result.exception
+    assert 'Chembur' in result.output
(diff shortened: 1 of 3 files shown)


tests/test_cli.py
```diff
@@ -868,3 +868,22 @@ def test_invoke_invalid_command(runner, tmpdir):
...
+        'Priority:High\n'
+    )
+
+    result = runner.invoke(cli, ['show', '1'])
+    assert 'High' in result.output
+
+
+def test_priority(runner):
+    result = runner.invoke(cli, [
+        'new', '-l', 'default', '--priority', 'Chembur', 'Event Test'
```
untitaker: Here you invoke `--priority` with the value `Chembur`, which is not a valid priority.

todoman/formatters.py
```diff
@@ -82,6 +82,8 @@ def detailed(self, todo):
             extra_rows += self._columnize('Description', todo.description)
         if todo.location:
             extra_rows += self._columnize('Location', todo.location)
+        if todo.priority:
```
untitaker: This is not necessary (and the corresponding test as well), because the priority is already shown as part of `self.compact(todo)`

todoman/cli.py
```diff
@@ -141,6 +141,10 @@ def validate_status(ctx=None, param=None, val=None):
 
 
 def _todo_property_options(command):
+    click.option(
+        '--priority', default='', callback=_validate_priority_param,
+        help=('Only show tasks with'
```
untitaker: This help text is wrong, we *set* the priority, not *filter* it

todoman/cli.py
```diff
@@ -141,6 +141,10 @@ def validate_status(ctx=None, param=None, val=None):
...
+              ' priority such as low, medium or high.'))(command)
```
hobarrera: I don't think `(command)` should be here.

untitaker: @hobarrera `click.option(..)` returns a decorator, and this is one way of applying it.

AnubhaAgrawal:
I make changes according to the suggestions given by @untitaker but again test cases are failed.

tests/test_cli.py
```diff
@@ -874,16 +874,16 @@ def test_show_priority(tmpdir, runner, create):
     create(
         'test.ics',
         'SUMMARY:harhar\n'
-        'Priority:High\n'
+        'Priority:high\n'
```
hobarrera: Priority here should be `1`. You should really use the `todo_factory` helper instead of `create`.

todoman/cli.py
```diff
@@ -141,6 +141,9 @@ def validate_status(ctx=None, param=None, val=None):
 
 
 def _todo_property_options(command):
+    click.option(
+        '--priority', default='', callback=_validate_priority_param,
+        help=('Only set the tasks with priority'))(command)
```
hobarrera: Not a very accurate help text. Use the two immediately below as a reference.

tests/test_cli.py
```diff
@@ -868,3 +868,22 @@ def test_invoke_invalid_command(runner, tmpdir):
     assert (
         'Error: Invalid setting for [main][default_command]' in result.output
     )
+
+
+def test_show_priority(tmpdir, runner, create):
+    create(
+        'test.ics',
+        'SUMMARY:harhar\n'
+        'Priority:high\n'
```
hobarrera: Looks like GitHub ate my previous comment.

Priority in files is numeric; should be `1` in this case. You should really use the `todo_factory` helper in this case anyway.

untitaker:
@hobarrera I see that you reviewed a testcase that doesn't test the new functionality.

@AnubhaAgrawal in https://github.com/pimutils/todoman/pull/242/commits/8bc088823cc72569c035cd330c3bdf06b5a6b09b you removed a testcase, but that was the more important one

tests/test_cli.py
```diff
@@ -868,3 +868,19 @@ def test_invoke_invalid_command(runner, tmpdir):
     assert (
         'Error: Invalid setting for [main][default_command]' in result.output
     )
+
+
+def test_show_priority(runner, todo_factory, todos):
+    todo = todo_factory(summary='harhar\n', priority=1)
+
+    result = runner.invoke(cli, ['show', '1'])
+    assert '1' in result.output
```
hobarrera: Output should say `high`, not `1`.

AnubhaAgrawal: if I write output as  ```high``` in place of ```1``` then this test case become failed.

hobarrera: The test passes because `1` is printed as the todo's ID, not priority.
Sorry, I mixed up a bit. In this case, `!!!` is printed to detonate the high priority, not `high`.

Sorry, my bad there! 😅

todoman/cli.py
hobarrera: Ugh, ignore me. I need to spot reviewing stuff right after I get up. :(

tests/test_cli.py
AnubhaAgrawal: Now, all test cases are passed except the one ```run_test(sort_key)```. I don't understand why it is failed.

hobarrera: I see those tests passing, but you have [style errors](https://travis-ci.org/pimutils/todoman/jobs/226445642#L215-L216).

You can run `tox -e flake8` to test these locally.

todoman/formatters.py
```diff
@@ -82,7 +82,6 @@ def detailed(self, todo):
             extra_rows += self._columnize('Description', todo.description)
         if todo.location:
             extra_rows += self._columnize('Location', todo.location)
-
```
hobarrera: I think removing this line was a mistake, everything else looks fine by me.

AnubhaAgrawal: by adding this line the test cases ```test_show_priority``` and ```test_priority``` are failed.

AnubhaAgrawal: Now, please reply what I should do??

hobarrera: An empty line shouldn't make a test fail (definitely not in this context). Are you sure you're not mistaken with something else?

AnubhaAgrawal: Now I don't understand where is a mistake? and when I check it locally it worked as we want. but why some test cases are failed I really don't understand.

AnubhaAgrawal: please reply, I think I do all requested changes Now, what change I should do because when I run this manually it work accordingly as we want.

hobarrera: This particular change is the removal of an empty line. It does not alter any functionality, and the change is completely unrelated to this PR.

todoman/formatters.py
```diff
@@ -82,7 +82,7 @@ def detailed(self, todo):
...
+
```
hobarrera: Please remove the newly-added empty spaces here. Spaces at the end of the line will make style checks fail.

todoman/cli.py
```diff
@@ -141,6 +141,9 @@ def validate_status(ctx=None, param=None, val=None):
 
 
 def _todo_property_options(command):
+    click.option(
+        '--priority', default='', callback=_validate_priority_param,
+        help=('The priority of the task is set.'))(command)
```
hobarrera: This is probably clearer written as `'The priority for this todo'`.

hobarrera:
Looks good, thanks!

@untitaker comments?

untitaker:
LGTM

hobarrera:
Thanks, @AnubhaAgrawal! 🎉

Issue #187, opened 2017-03-09
Provide autocompletion for bash and zsh


WhyNotHugo:

stweller:
Hi @hobarrera,

I do have a basic version of tab completion for bash (can also be used in zsh with the compatibility layer, of course). I can open a pull request, but I need to know where to actually put the file in the project. Should we have a new folder `completion` or something?

hobarrera:
Hey, thanks!

A new `contrib` directory would've fine. If you can, installation instructions for these would be nice (though I write those if you've no idea how).

stweller:
I opened a pull request just now. Let's continue the discussion there.

johnjameswhitman:
Recently I've been working on trying to get completion working with fish, and realized that `click` has built-in support to generate completions for a variety of shells (see #543 for a little more context):

1. [`click` docs on generating completions for a given shell](https://click.palletsprojects.com/en/8.1.x/shell-completion/#enabling-completion)
2. [`click` docs on customizing argument and option value completions](https://click.palletsprojects.com/en/8.1.x/shell-completion/#overriding-value-completion)

Right now you can get (1) out of the box w/ `todo`; however, (2) requires implementing a `shell_complete` method per argument or option. It's not perfect, but I've got a sample for `todo new --list` in this fork:
https://github.com/johnjameswhitman/todoman/blob/4a00d7ecb20028f4d8f3173928e4cd24849e97c4/todoman/cli.py#L350-L370

Unfortunately, it doesn't seem like you can access the `AppContext` that `cli` sets up, so you have to create it within the shell completion function. This also means you don't have access to the `config` argument passed into the main cli command (at least, I can't figure out how to find it after some debugging). It seems like the completions in `contrib` also have this limitation, and either parse the default config location ([zsh](https://github.com/pimutils/todoman/blob/31a96d414e2e685be90bfe3c7f067934252d6278/contrib/completion/zsh/_todo#L50-L57)) or run todo w/o specifying a config ([bash](https://github.com/pimutils/todoman/blob/31a96d414e2e685be90bfe3c7f067934252d6278/contrib/completion/bash/_todo#L7-L11)). So, implementing shell completion within `todoman/cli.py` may make it easier to maintain across both bash and zsh, plus enable fish.

Is adding click-based support for value-completions a contribution the project would welcome?

WhyNotHugo:
> Is adding click-based support for value-completions a contribution the project would welcome?

[...]

PR #260, opened 2017-07-24
Add initial support for (bash) autocompletion.

stweller:
This should provide part of the fix for #187 (i.e. autocompletion for bash, but not yet for zsh).

Note that the completion is at the moment completely static. For the future, it would be great to add dynamic completion, e.g. if you type ``todo list --location <Tab>``, you would get all existing locations as completion options. However at the moment that would require parsing json from a bash script which is not so nice. Maybe we can add a ``--format`` switch to ``todo list`` in the future, so it is easy to list e.g. all locations from a bash script.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -21,5 +21,6 @@ Authors are listed in alphabetical order.
 * Paweł Fertyk <pfertyk@openmailbox.org>
 * Rimsha Khan <rimshakhan.rk03@gmail.com>
 * Sakshi Saraswat <saraswatsakshi.121@gmail.com>
+* Stephan Weller <stephan.weller@web.de>
 * Swati Garg <swati4star@gmail.com>
 * Thomas Glanzmann <thomas@glanzmann.de>
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,11 @@ Changelog
 This file contains a brief summary of new features and dependency changes or
 releases, in reverse chronological order.
 
+v3.2.2
+------
+
+* Initial support for (bash) autocompletion.
+
 v3.2.1
 ------
 
--- a/contrib/completion/bash/_todo
+++ b/contrib/completion/bash/_todo
@@ -1,0 +1,94 @@
+_todo_single_dash_options()
+{
+	local prev_word=$1
+
+	case $prev_word in
+		copy)
+			echo "-l"
+			;;
+		edit)
+			echo "-s -d -i"
+			;;
+		list)
+			echo "-s"
+			;;
+		move)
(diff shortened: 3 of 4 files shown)


stweller:
I don't know whether we could and should add some instructions to ``setup.py`` to install the completion function automatically. At the moment, everything needs to be installed manually.

hobarrera:
Hi!

First of all, sorry for taking so long to reply. Looks like GitHub either didn't send me the notification for this, or I completely oversaw it.

Secondly, thanks for contributing! 😃 

I'm not seeing the expected results though (nor bash, nor zsh-with-bash-completion):

```
$ todo <TAB>
cancel  copy    delete  done    edit    flush   list    move    new     show    
$ todo list <TAB>
cancel  copy    delete  done    edit    flush   list    move    new     show    
$ todo list 
```

Apparently, something's not totally working with the inner `case`.

stweller:
Hi @hobarrera ,

well, as I mentioned above, at the moment the completion is completely static. What would you expect to see after the second `<Tab>`? To see e.g. all existing lists, I would need to extract those somehow from todoman, and that would only be possible by parsing json from a bash script, which is really awkward. If there was some option to list all lists, all locations etc., it could be implemented pretty easily.

BTW: Options should work, so after `todo list -<Tab>` you should get something sensible.

hobarrera:
Oh, right, I get it. I'd expected to `todo list <Tab>` to output the same as `todo list -<Tab>`.

This PR looks good, but would you mind adding yourself to `AUTHORS` and updating the changelog (next version would be `3.2.2`).

In case you'd like work on the todo `id`s, [`jq`](https://stedolan.github.io/jq/) would probably help.

Again, thanks!

stweller:
I updated the changelog and authors file. As for jq, that would probably work, but I don't know whether it's worthwhile to add a dependency to an otherwise unrelated project just for autocompletion. I think a switch like --format for a user-defined output format would be better. Maybe I can come up with a PR for something like that later on.

hobarrera:
Thanks!

> I think a switch like --format for a user-defined output format would be better.

I'd rather not implement custom user formatting -- the whole idea of `--porcelain` is to avoid having very complex format configuration scenarios, and just allow users to use wrappers for these things.

`sed` or `grep` might work, since things like `^\s+"id": (\d+),` (this matches the id) can be used.

    {
        "completed": false,
        "due": 1505012400,
        "id": 72,
        "list": "personal",
        "percent": 0,
        "priority": 1,
        "summary": "Blah"
    },

stweller:
Well, '\d' is a perl extention, that would only work with `grep -P`, but ok, I'll see whether I can do something.

PR #261, opened 2017-07-25
Spelling manpage

legrostdg:


hobarrera:
Thanks! Looks like GitHubs emails have been failing recently and I've missed a bunch of stuff in many projects: 😐

PR #264, opened 2017-08-25
Fix start-up crash and drop click_log

hobarrera:
Fix a crash caused by a change of interface in `click_log`.
Given that the new interface is a bit more verbose, and doesn't really add much to using base `logging`, I'll adhere with the latter.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,12 @@ Changelog
 This file contains a brief summary of new features and dependency changes or
 releases, in reverse chronological order.
 
+v3.2.1
+------
+
+* Fix start-up crash caused by click_log interface change.
+* Dropped runtime dependency: ``click_log``.
+
 v3.2.0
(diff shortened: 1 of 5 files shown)


untitaker:
Well it does provide color formatting :)

hobarrera:
I tried upgrading to `click_log` v0.2.0, but it doesn't seem to work.

Running `todo -v DEBUG` results in no log output. Looks like `click_log` only changes the level for the current module's logger (`todoman.cli` in this case), and won't affect other modules.

hobarrera:
@untitaker I'm very open to re-including `click-log` in future (a few extra things needs polishing, IMHO), but for now, I'm merging this since it fixes a critical, fatal bug that doesn't allow the app to run.

I've opened https://github.com/click-contrib/click-log/issues/9 with one of the issues I've encountered. If I've the time, I might try to look into it.

PR #268, opened 2017-08-31
Make bash autocompletion dynamic (#187).

stweller:
OK, here is a more dynamic version. It can complete task id's and lists. In theory it could also complete locations, however location is not present in "porcellain" output. I think it should definitely be there, should I open an issue for that?

### DIFF
--- a/contrib/completion/bash/_todo
+++ b/contrib/completion/bash/_todo
@@ -1,3 +1,21 @@
+_todo_get_ids()
+{
+	local pattern='^\s*"id": \([0-9]\+\),'
+	todo --porcelain list | grep -o "${pattern}" | sed -e "s/${pattern}/\1/" | sort
+}
+
+_todo_get_lists()
+{
+	local pattern='^\s*"list": "\(.*\)",'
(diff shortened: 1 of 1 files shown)


contrib/completion/bash/_todo
```diff
@@ -1,3 +1,21 @@
+_todo_get_ids()
+{
+	local pattern='^\s*"id": \([0-9]\+\),'
+	todo --porcelain list | grep -o "${pattern}" | sed -e "s/${pattern}/\1/" | sort | uniq
```
hobarrera: Is `uniq` necesary here?

hobarrera:
Looks like your email in your commits doesn't match an email on your github account, so your profile isn't linked it commit authorship pages, etc.
Dunno if you care about that, but just pointing that out (note: I completely don't mind though 🙂 )

contrib/completion/bash/_todo
stweller: No, you're right, that can be omitted here. `id` should be unique anyway.

stweller:
I updated my github profile btw, so the commits should now be linked correctly.

hobarrera:
Thanks!

PR #271, opened 2017-08-31
Location fixes

stweller:
Well, that wasn't particularly hard...


stweller:
Should be fixed now. BTW: I had some trouble finding out which prerequisites I need to install to run the tests locally. Maybe you want to add a paragraph to the README.

hobarrera:
Thanks!

I've addresses improving those docs as part of #275.

PR #279, opened 2017-09-10
Explicitly define which PyPI endpoint to push to

hobarrera:
Pushes have been failing lately (and not just taking a few hours/days to
finish as they used to).

It's most likely due to this:
* https://mail.python.org/pipermail/distutils-sig/2017-June/030766.html
* https://packaging.python.org/guides/migrating-to-pypi-org/#uploading
* https://github.com/pypa/warehouse/issues/2342

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,6 +69,7 @@ deploy:
     on:
       tags: true
     distributions: "sdist bdist_wheel"
+    server: https://upload.pypi.org/legacy/
   - provider: releases
     api_key:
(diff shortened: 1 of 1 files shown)


untitaker:
Couldn't you start using twine?

hobarrera:
That's what I was originally using (pre-github). It'll be my next approach if travis' built-in deploy starts failing again, but this seems simpler consider we also deploy as github releases.

hobarrera:
None of the jobs [here](https://travis-ci.org/pimutils/todoman/builds/274331490) show any output for deploy.

I'm not sure if travis even did anything or not. I'll move over to twine.

hobarrera:
Now using twine (see 5dc49b2dae9504c5c9c16a411420ef471f427352).
Thanks to all my failed attempts, we can be sure that any failures are shown in Travis' logs too, so there should be a bit less mystery about missing releases.

@untitaker Have you considered doing similar CI-based releases for vdirsyncer?

PR #280, opened 2017-09-26
Reintroduce click-log

WhyNotHugo:
Looks like passing the root logger could have solved our issue with it last time. This is also now the default too. 🎉

### DIFF
--- a/todoman/cli.py
+++ b/todoman/cli.py
@@ -197,7 +197,7 @@ _interactive_option = click.option(
 
 
 @click.group(invoke_without_command=True)
-@click_log.simple_verbosity_option(None)
+@click_log.simple_verbosity_option()
 @click.option('--colour', '--color', default=None,
               type=click.Choice(['always', 'auto', 'never']),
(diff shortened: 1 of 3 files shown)


WhyNotHugo:
Any comment son this, @untitaker (I'm asking just in case you think I missed something).

todoman/cli.py
```diff
@@ -194,13 +197,7 @@ def formatter(self):
...
 
 @click.group(invoke_without_command=True)
-@click.option(
-    '-v',
-    '--verbosity',
-    type=click.Choice(['CRITICAL', 'ERROR', 'WARNING', 'INFO', 'DEBUG']),
-    help='Set verbosity to the given level.',
-    default='WARNING',
-)
+@click_log.simple_verbosity_option(None)
```
untitaker: Explicit parameter should not be necessary now, right?

untitaker:
Lgtm

todoman/cli.py
WhyNotHugo: Good catch, thanks!

PR #281, opened 2017-10-01
Reformat using yapf, and add it to CI

WhyNotHugo:
[yapf](https://github.com/google/yapf) is a code auto-formatting tool. I've configured it to strictly follow PEP-8, meaning it's another pep-8 checker, but the present configuration also allows quickly reformatting non-compliant code.

I'm generally happy with the changes done (it seems to introduce consistency, and clean up thing I generally didn't like).

I'm just hesitant because of [this bug](https://github.com/google/yapf/issues/462) which does affect the  codebase, and I'd also prefer to see [this](https://github.com/google/yapf/issues/461) implemented (though I guess that's not as critical).

Some assertions also [look kinda odd](https://github.com/pimutils/todoman/compare/master...yapf#diff-5d99b6fbb067a47380e72c50ef6357d1R236), though that *does* seem to be the _"proper"_ style.

My final hesitation is the size of this PR, since it kinda degrades the usefulness of things like `git blame` to find out why we introduced certain changes (though it's possible that the codebase is small and compact enough that this isn't really an issue).

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -30,7 +30,7 @@ config = ConfigObj(
     None,
     configspec=specpath,
     stringify=False,
-    list_values=False
+    list_values=False,
 )
 validator = validate.Validator()
 config.validate(validator)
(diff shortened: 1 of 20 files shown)


untitaker:
For now I use the combination of flake8 + plugins and autopep8 in vdirsyncer... I didn't hear particularly good things about yapf. I'm interested in how it will work out for you.

PR #290, opened 2017-11-07
Option to read new description from stdin

untitaker:
Will add tests if you approve of API

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -873,3 +873,11 @@ def test_edit_raw(todo_factory, runner):
 
     assert not result.exception
     assert not result.output
+
+
+def test_new_description_from_stdin(runner):
+    result = runner.invoke(cli, ['new', '-l', 'default', '-r', 'hello'], input='world\n')
+    assert not result.exception
+    result = runner.invoke(cli, ['show', '1'], input='world\n')
+    assert not result.exception
+    assert 'world' in result.output
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -875,10 +875,10 @@ def test_edit_raw(todo_factory, runner):
     assert not result.output
 
 
-def test_new_description_from_stdin(runner):
+def test_new_description_from_stdin(runner, todos):
     result = runner.invoke(cli, ['new', '-l', 'default', '-r', 'hello'],
                            input='world\n')
     assert not result.exception
-    result = runner.invoke(cli, ['show', '1'], input='world\n')
-    assert not result.exception
-    assert 'world' in result.output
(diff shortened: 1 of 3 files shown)


untitaker:
@WhyNotHugo any opinion on this? I use this feature everyday

todoman/cli.py
```diff
@@ -318,11 +318,18 @@ def invoke_command(click_ctx, command):
     callback=_validate_list_param,
     help='The list to create the task in.'
 )
+@click.option(
+    '--read-description',
+    '-r',
+    is_flag=True,
+    default=False,
+    help='Read task description from stdin'
```
WhyNotHugo: Please finish sentences with `.` in help texts.

tests/helpers.py
```diff
@@ -1,9 +1,9 @@
+import pytest
```
WhyNotHugo: The was actually right before:

https://www.python.org/dev/peps/pep-0008/#module-level-dunder-names

tests/test_cli.py
```diff
@@ -873,3 +873,12 @@ def test_edit_raw(todo_factory, runner):
 
     assert not result.exception
     assert not result.output
+
+
+def test_new_description_from_stdin(runner):
+    result = runner.invoke(cli, ['new', '-l', 'default', '-r', 'hello'],
+                           input='world\n')
+    assert not result.exception
```
WhyNotHugo: I believe it's also important to check that the word `world` ends up in the right field in the created todo. Probably can be done following the pattern [here](https://github.com/pimutils/todoman/blob/05b7f915be7b04fb16a0c1fb5940d5f504cc7add/tests/test_cli.py#L709-L710).

WhyNotHugo:
Don't forget about the CHANGELOG entries 😉

tests/helpers.py
untitaker: I fixed it because Flake8 was complaining.

untitaker:
I think everything is fixed now. The stylecheck will fail because the import order is "wrong" according to flake8 though.

tests/test_cli.py
```diff
@@ -873,3 +873,12 @@ def test_edit_raw(todo_factory, runner):
 
     assert not result.exception
     assert not result.output
+
+
+def test_new_description_from_stdin(runner, todos):
+    result = runner.invoke(cli, ['new', '-l', 'default', '-r', 'hello'],
+                           input='world\n')
+    assert not result.exception
+    todo, = todos()
```
WhyNotHugo: Nice, hand't thought of using `todo,`, a lot cleaner than my approach! 😉

PR #291, opened 2017-11-09
Case-insensitive list name matching

wichmannpas:
This patch brings case-insensitive list name matching for the list parameter if all list names are unique when converted to lowercase (i.e., if the case-insensitive matching would yield conflicts for any list, it is not used).

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -11,7 +11,7 @@ from dateutil.tz import tzlocal
 from freezegun import freeze_time
 from hypothesis import given
 
-from tests.helpers import pyicu_sensitive
+from tests.helpers import fs_case_sensitive, pyicu_sensitive
 from todoman.cli import cli, exceptions
 from todoman.model import Database, Todo
 
@@ -66,6 +66,7 @@ def test_percent(tmpdir, runner, create):
     assert '78%' in result.output
 
 
+@fs_case_sensitive
 @pytest.mark.parametrize(
     'list_name', [
         'default',
@@ -77,6 +78,7 @@ def test_list_case_insensitive(tmpdir, runner, create, list_name):
     assert not result.exception
 
 
+@fs_case_sensitive
 def test_list_case_insensitive_collision(tmpdir, runner, create):
     """
     Test that the case-insensitive list name matching is not used if
@@ -94,6 +96,7 @@ def test_list_case_insensitive_collision(tmpdir, runner, create):
     assert not result.exception
 
 
+@fs_case_sensitive
 def test_list_case_insensitive_other_collision(tmpdir, runner, create):
     """
     Test that the case-insensitive list name matching is used if a
@@ -109,15 +112,14 @@ def test_list_case_insensitive_other_collision(tmpdir, runner, create):
     assert not result.exception
 
 
-@pytest.mark.parametrize(
-    'list_name', [
-        'nonexistant',
-        'NONexistant',
-    ])
-def test_list_inexistant(tmpdir, runner, create, list_name):
-    result = runner.invoke(cli, ['list', list_name])
+def test_list_inexistant(tmpdir, runner, create):
+    result = runner.invoke(cli, ['list', 'nonexistant'])
     assert result.exception
-    assert 'Error: Invalid value for "lists": %s' % list_name in result.output
+    assert 'Error: Invalid value for "lists": nonexistant' in result.output
+
+    result = runner.invoke(cli, ['list', 'NONexistant'])
+    assert result.exception
+    assert 'Error: Invalid value for "lists": NONexistant' in result.output
 
 
 def test_show_existing(tmpdir, runner, create):
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -66,10 +66,58 @@ def test_percent(tmpdir, runner, create):
     assert '78%' in result.output
 
 
-def test_list_inexistant(tmpdir, runner, create):
-    result = runner.invoke(cli, ['list', 'nonexistant'])
+@pytest.mark.parametrize(
+    'list_name', [
+        'default',
+        'DEfault',
+        'deFAUlT',
+    ])
+def test_list_case_insensitive(tmpdir, runner, create, list_name):
+    result = runner.invoke(cli, ['list', list_name])
+    assert not result.exception
+
+
+def test_list_case_insensitive_collision(tmpdir, runner, create):
+    """
+    Test that the case-insensitive list name matching is not used if
+    colliding list names exist.
+    """
+    tmpdir.mkdir('DEFaUlT')
+
+    result = runner.invoke(cli, ['list', 'deFaulT'])
     assert result.exception
-    assert 'Error: Invalid value for "lists":' in result.output
+
+    result = runner.invoke(cli, ['list', 'default'])
+    assert not result.exception
+
+    result = runner.invoke(cli, ['list', 'DEFaUlT'])
+    assert not result.exception
+
+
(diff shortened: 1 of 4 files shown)


todoman/cli.py
```diff
@@ -53,14 +53,24 @@ def _validate_list_param(ctx, param=None, name=None):
...
-        )
+    lists = {
+        l.name: l
+        for l in ctx.db.lists()
+    }
+    normalized_lists = {
+        l.name.lower(): l
+        for l in lists.values()
+    }
+    if len(lists) == len(normalized_lists):
```
WhyNotHugo: Well though 👍

todoman/cli.py
```diff
@@ -53,14 +53,24 @@ def _validate_list_param(ctx, param=None, name=None):
...
+        # all normalized list names are unique, use case-insensitive comparison
+        lists = normalized_lists
+    for n in lists:
```
WhyNotHugo: I'm not sure this `for` is correct. Won't this just return the first list from those provided, rather than all of them (this never seems to return a list).

todoman/cli.py
```diff
@@ -53,14 +53,24 @@ def _validate_list_param(ctx, param=None, name=None):
...
-    for l in ctx.db.lists():
-        if l.name == name:
-            return l
-    else:
-        list_names = [l.name for l in ctx.db.lists()]
-        raise click.BadParameter(
-            "{}. Available lists are: {}".format(name, ', '.join(list_names))
-        )
+    lists = {
+        l.name: l
```
WhyNotHugo: `l` is an ambiguous variable name, and flake will complain about it. Please use something like `list_`.

wichmannpas:
Thanks for your review. I have updated the PR accordingly.
Furthermore, I have improved the feedback to display the original list name entered by the user (i.e., without normalization to lowercase).

todoman/cli.py
```diff
@@ -53,14 +53,25 @@ def _validate_list_param(ctx, param=None, name=None):
...
+    lists = {
+        list_.name: list_
+        for list_ in ctx.db.lists()
+    }
+    normalized_lists = {
+        list_.name.lower(): list_
+        for list_ in lists.values()
+    }
+    original_name = name
+    if len(lists) == len(normalized_lists):
```
untitaker: Can't we use normalized lists in more situations? Imagine the following:

* We have three lists: `a`, `A`, `b`
* The user selects the list `B`

In this situation not *all* list names are unique, but the one *we care about* is.

tests/test_cli.py
```diff
@@ -66,10 +66,25 @@ def test_percent(tmpdir, runner, create):
     assert '78%' in result.output
 
 
+def test_list_case_insensitive(tmpdir, runner, create):
+    result = runner.invoke(cli, ['list', 'default'])
```
untitaker: You can use http://doc.pytest.org/en/latest/example/parametrize.html here

todoman/cli.py
WhyNotHugo: Good catch, I hadn't really though about that. I guess you'd have to count how many lists normalize the the same as the input one, and make sure that's `== 1`.

tests/test_cli.py
```diff
@@ -66,10 +66,25 @@ def test_percent(tmpdir, runner, create):
...
+    assert not result.exception
+
+    result = runner.invoke(cli, ['list', 'DEfault'])
+    assert not result.exception
+
+    result = runner.invoke(cli, ['list', 'deFAUlT'])
+    assert not result.exception
```
WhyNotHugo: These tests look good. Can you also add one for the failure scenario (eg: lists `default` and `DEFAULT` both exist)?

wichmannpas:
Thanks for your comments. I have updated the logic to skip the normalization of all lists, but instead only look for fuzzy matches of the requested list.
In addition, I have added more tests for that scenario as well as using the parametrize decorator as suggested.

wichmannpas:
Hrhr. It seems that the parametrization using parameters that are identical except for their case breaks on OS X (which is rather comprehensible for it having a case-insensitive file system by default).
Any ideas on how or where to fix that @untitaker?

WhyNotHugo:
We can probably attempt to detect if the filesystem is case-insensitive before running the test.

Check how the [`pyicu_sensitive`](https://github.com/pimutils/todoman/blob/f1e3221ef4c1d78b004699d3fc58768d41e67dc6/tests/helpers.py) decorator is defined. You can probably do something similar like:

```python
def is_fs_case_sensitive():
    try:
        create_dir('test')
        create_dir('TEST')
        return True
    except FileExists:
        return False

case_sensitive = pytest.mark.skipif(
    not is_fs_case_sensitive(),
    reason="This test cannot be run on case insensitive systems.",
)
```

wichmannpas:
Ok, I have added such a helper. The tests seem to pass for OS X now as well.
For the test_list_inexistant, I have removed the parametrization again as I think that test should run on OS X as well.

untitaker:
This looks fine to me, but I think those special decorators might as well be fixtures:

`def test_foo(requires_case_sensitive_fs): ...`

Then you can just use the `tmpdir` fixture inside, and call `pytest.skip` when you want to skip the test.

It would also be faster to just skip that test if on OS X, without any real feature test. But I guess I see why you did it that way.

wichmannpas:
> This looks fine to me, but I think those special decorators might as well be fixtures:
> 
> def test_foo(requires_case_sensitive_fs): ...
> 
> Then you can just use the tmpdir fixture inside, and call pytest.skip when you want to skip the test.

That is a good point. However, that would probably not work as the test needs to be marked for skipping before parametrize tries to create the temp directories for the tests failing on unsupported platforms.

untitaker:
The skipping would be done inside the fixture function body, which is executed before your test.

wichmannpas:
I have updated the test to use a fixture. Is this the way you suggest?

tests/conftest.py
```diff
@@ -160,6 +160,15 @@ def inner(**filters):
     return inner
 
 
+@pytest.fixture
+def requires_case_sensitive_fs(tmpdir):
+    try:
+        tmpdir.mkdir('test')
```
untitaker: Please rename this to something more unique since it might conflict with actual test data otherwise. 

Also I noticed the first mkdir should be outside of the try-catch.

untitaker:
Yes exactly. Please fix the one remaining comment then we can merge this :)

tests/conftest.py
wichmannpas: Good point. I have updated the commit.

wichmannpas:
Hrm, the tests seem to be failing on OS X again now. It seems that the tmpdir fixture tries to create the temporary directories before the requires_case_sensitive_fs fixture skips the test.

untitaker:
In Travis it seems that py.test raises a different kind of error than you
catch.

untitaker:
To be clear I can't test anything on OS X locally... the fixture thing above was more meant to be a suggestion. If it turns out that it's too much effort to do it this way or if @WhyNotHugo objects then we can merge it in its prev. state :)

wichmannpas:
Alright then, I have updated the commit to catch py.error.EEXIST instead. Let's see whether that fixes the problem and otherwise go back to the state without the fixture.

untitaker:
Ok scratch that, let's go with your first solution. Sorry for the noise!

wichmannpas:
yep, seems that it cannot reliably assumed that the fixture is run before the parametrization. I have returned to the previous version.

untitaker:
Sorry for the wait, merged!

PR #294, opened 2018-01-14
Add valid fields for --sort

### LINKED ISSUE #293: Valid fields for `--sort` are undocumented
We don't list the options valid for `--sort` anywhere, and the only way to find out what they are is inspecting the code.

rimshaakhan:
Fixes #293 
#### Changes made:
- Add important fields in help text of --sort argument
- Add all the fields of -- sort argument in usage.rst

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -71,3 +71,24 @@ The format is JSON, with a single array containing each todo as a single entry
 given field, it will be printed as ``null``.
 
 Fields MAY be added in future, but will never be removed.
+
+Sorting
+-------
+
+The tasks can be sorted with the ``--sort`` argument. Sorting may be done according to the following fields:
+    - ``description``
+    - ``location``
+    - ``status``
+    - ``summary``
+    - ``uid``
+    - ``rrule``
+    - ``percent_complete``
+    - ``priority``
+    - ``sequence``
+    - ``categories``
+    - ``completed_at``
+    - ``created_at``
+    - ``dtstamp``
+    - ``start``
+    - ``due``
+    - ``last_modified``
--- a/todoman/cli.py
+++ b/todoman/cli.py
@@ -532,7 +532,12 @@ def move(ctx, list, ids):
 @click.option('--grep', help='Only show tasks with message containg TEXT')
 @click.option(
     '--sort',
-    help='Sort tasks using these fields',
+    help=(
+          'Sort tasks using fields like : '
+          '"start", "due", "priority", "created_at", "percent_complete" etc.\n '
+          ' For all fields click: '
+          '<https://todoman.readthedocs.io/en/stable/usage.html> '
+          ),
     callback=_sort_callback,
 )
 @click.option(


docs/source/usage.rst
```diff
@@ -71,3 +71,35 @@ The format is JSON, with a single array containing each todo as a single entry
 given field, it will be printed as ``null``.
 
 Fields MAY be added in future, but will never be removed.
+
+Sorting
+-------
+
+The tasks can be sorted with the --sort argument. Sorting may be done according to the following fields:
```
untitaker: Please use <code>\`\`</code> around "literal text" such as option names or field names. Otherwise they may be wrongly rendered.

docs/source/usage.rst
```diff
@@ -71,3 +71,35 @@ The format is JSON, with a single array containing each todo as a single entry
...
+
+- STRING_FIELDS
+    * description
+    * location
+    * status
+    * summary
+    * uid
+    * rrule
+
+- INT_FIELDS
```
untitaker: I don't think it's necessary to have those categories, they're just for internal use.

todoman/cli.py
```diff
@@ -532,7 +532,8 @@ def move(ctx, list, ids):
 @click.option('--grep', help='Only show tasks with message containg TEXT')
 @click.option(
     '--sort',
-    help='Sort tasks using these fields',
+    help=('Sort tasks using fields like "start", "due", "priority", "created_at", "percent_complete" etc.'
+          '(for all fields see todoman/docs/source/usage.rst)'),
```
untitaker: I would use the link `https://todoman.readthedocs.io/en/stable/usage.html`

docs/source/usage.rst
```diff
@@ -71,3 +71,35 @@ The format is JSON, with a single array containing each todo as a single entry
...
+    * priority
+    * sequence
+
+- LIST_FIELDS
+    * categories
+
+- DATETIME_FIELDS
+    * completed_at
+    * created_at
+    * dtstamp
```
untitaker: Same here, please use <code>\`\`</code>, e.g. <code>\`\`completed_at\`\`</code>. See also http://docutils.sourceforge.net/docs/user/rst/quickref.html#inline-markup

docs/source/usage.rst
untitaker: I just meant that you don't have to separate the attributes into `INT_FIELDS`, `STRING_FIELDS`, `DATETIME_FIELDS` etc because it doesn't make a difference to the user for this usecase.

There's nothing wrong with sorting by `uid`... I think todoman doesn't fail silently but actually tries to sort by the value of the UID property. If it makes sense to do that is a different thing, but who are we to judge the user :)

untitaker:
Don't worry about the test failures, they are unrelated.

docs/source/usage.rst
RJ722: Yes, indeed we cannot judge the user! ;-)

But, I think todoman is failing silently if we try and sort by `uid`.
Here's the output of `todo list --sort uid` on my machine:
```
42  [ ]  !!   07-07-2017 00:37  ---------------------
36  [ ]       23-06-2017 15:29  ---------------------
54  [ ]  !    21-11-2017 13:52  ---------------------
46  [ ]  !    30-06-2017 23:44  ---------------------
53  [ ]  !!!  23-10-2017 13:45  ---------------------
```
I also tried something similar with `location` (`STRING_FIELDS`), I made three todos whose location started with `a`, `b` and `c`, after sorting they were displayed in the following order: `b`, `a` and `c`.     ¯\_(ツ)_/¯

I haven't tried `LIST_FIELDS` and `RRULE_FIELDS` because I'm not sure of how todoman should ideally behave?

untitaker: I can reproduce this bug, let's discuss this in a separate issue.

untitaker: https://github.com/pimutils/todoman/issues/295

todoman/cli.py
WhyNotHugo: I think we can also programatically generate a list of sortable fields.
But it's fine as-is right now, we can improve on that later. 🙂

WhyNotHugo:
Thanks @rimshaakhan for this, and @untitaker for looking into it too. 🙂 

The `python3.4` failure looks unrelated to your code. It's actually the `pyicu` test, which installs `pyicu` (an library with which we used to have some conflicts).  
I'll look into it separately, master is probably affected too.

todoman/cli.py
```diff
@@ -532,7 +532,12 @@ def move(ctx, list, ids):
 @click.option('--grep', help='Only show tasks with message containg TEXT')
 @click.option(
     '--sort',
-    help='Sort tasks using these fields',
+    help=(
+          'Sort tasks using fields like : '
+          '"start", "due", "priority", "created_at", "percent_complete" etc.\n '
+          ' For all fields click: '
```
RJ722: IMHO 'For all fields ~~click~~ visit:' would sound better.

rimshaakhan:
@WhyNotHugo  @untitaker  Can we please merge the PR?

untitaker:
Sorry for the delay, thanks!

rimshaakhan:
Thank you! :)

WhyNotHugo:
Thanks!

PR #308, opened 2018-04-10
Skip test that crashes due to invalid input

WhyNotHugo:
Fix a test that fails since todoman crashes attempting to process and invalid ICS file.

Actually just skip this scenario, since crashing when processing broken input is probably the best we can do (and what we do in almost any other similar scenario).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,4 +62,5 @@ jobs:
 
 matrix:
   allow_failures:
+    - python: pypy3.3-5.2-alpha1
     - python: 3.7-dev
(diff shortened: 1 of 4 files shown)


WhyNotHugo:
This fixes master, which is once again broken due to upstream changes (though upstream didn't actually break stuff this time).

I've also set travis to run on a daily basis, if no there have been no runs for 24 hours. This should help catch these things faster.

WhyNotHugo:
Triggering CI, which got stuck.

WhyNotHugo:
Looks like there's a slight [coverage drop](https://codecov.io/gh/pimutils/todoman/pull/308/changes) that should be looked into. Not sure if it's now-dead code, or if it's only reached with dirty input and needs to be re-evaluated.

Issue #292, opened 2018-01-05
Allow custom editor


varac:
It would be nice to edit a new/existing item using my editor of choice, like [khard](https://github.com/scheibler/khard) does.
Meanwhile, where can I find a documentaion on the current integrated editor ?
Especially when removing a long description from an item I feel the pain of a different, unknown editor.

untitaker:
Hit F1 for a list of shortcuts. One of them will open EDITOR with the current textfield content

varac:
Thx, I was not aware of that. However, this opens an editor for a particular text field. khard uses an external editor to edit all fields at once, which would be my favorite.

WhyNotHugo:
The latest version lets you do `todo edit --raw`, but you really need to know what you're doing with that (since we don't validate modifications).
Is that kinda what you're looking for?

varac:
Looks good, but I'm hesitant to use it when you say you don't validate. Also, `--raw` don't exist for `todo new`, which is unfortunate.

WhyNotHugo:
> new with raw would be the same as you just creating a file yourself, no?

I guess so. Though it might be a bit better to create a blank `todo` and allow editing that (eg: with a created date, uid, etc).

OTOH, I guess we _could_ try to validate todos edited with `raw`, by:

* Copy the file to $TMP.
* Open it with the user's editor.
* Try and parse the file:
  * If that works, copy it back.
  * If there's an error, show it, and offer to reopen the editor.

Thoughts?

varac:
@untitaker, @WhyNotHugo:

>>   new with raw would be the same as you just creating a file yourself, no?

> I guess so.

No. Pls have a look how `khard` does it. `khard new` starts your configured editor, seeds it with all the available fields and (hopefully) validates it on saving.
If we consider this option, we should make it ppl as easy as possible, without the need to remember field names i.e.

WhyNotHugo:
FWIW, what fields are you trying to edit that you _can't_ edit via the UI?
I'd rather improve the UI, instead of converting to a custom format for the editor.

varac:
@WhyNotHugo I can edit all fields using the UI, it's just that it would be much more convinient to use the keybindings of my favorite editor. So after all, not super important but rather a nice-to-have, and I can understand if you won't put a lot of effort into it. Nonetheless, maybe somebody else likes to implement it ?
My goal would be to have all related cli tools (khard, khal, todoman) usable with your favorite editor, but I can live with the current situation.

[...]

PR #312, opened 2018-05-19
Add "todo edit --description" for comfy editing in external editor

hut:
My primary use case of todoman is comfortable editing of descriptions of tasks in a separate editor. The reason is that most data of a task is simple enough to be editable via low-bandwidth interfaces like android clients or thunderbird/lightning. But the description may contain a long text, so the use of a dedicated text editor is warranted.

Currently the same thing can be achieved by running `todo edit <taskid>`, pressing the arrow key down, then pressing `<C-O>`. With the option from this PR, all you need is to pass the `--description` option to the `edit` command to open the item directly in the editor.

Feel free to change the name of the option. Related issue: #292

### DIFF
--- a/todoman/cli.py
+++ b/todoman/cli.py
@@ -383,11 +383,18 @@ def new(ctx, summary, list, todo_properties, read_description, interactive):
         "Only use this if you REALLY know what you're doing!"
     )
 )
+@click.option(
+    '--description',
+    is_flag=True,
+    help=(
+        'Edit the description in an external editor.'
+    )
+)
 @_todo_property_options
 @_interactive_option
 @with_id_arg
 @catch_errors
-def edit(ctx, id, todo_properties, interactive, raw):
+def edit(ctx, id, todo_properties, interactive, raw, description):
     '''
     Edit the task with id ID.
     '''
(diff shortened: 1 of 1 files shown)


hut:
I'm not sure what I should do (or whether i should do anything) about codecov check failure.

WhyNotHugo:
Thanks, this looks pretty good. `--description` sounds _slightly_ odd, but I can't think of anything better, so it's fine by me (@untitaker, any comments?).

Can you please add some tests for that cover this new feature?
Also, please add yourself to AUTHORS and include a CHANGELOG entry (this'd be v3.5.0)

Again, thanks!

untitaker:
This is inconsistent with `todo new` which has an option for reading the
description from stdin.

Not entirely sure what's better but I think we should standardize on this.

hut:
Is there some recommended way to use `todo new --read-description` combined with a text editor?

untitaker:
You can use -i. There is a shortcut to open $EDITOR for the currently focused textfield in the interactive UI

Or you write a wrapper script

On 19 May 2018 21:55:34 GMT+02:00, hut <notifications@github.com> wrote:
>Is there some recommended way to use `todo new --read-description`
>combined with a text editor?
>
>-- 
>https://github.com/pimutils/todoman/pull/312#issuecomment-390428681

-- 

hut:
I guess it's best to add an option that reads from stdin. While using the text editor directly is probably the most common use case... Once you got another use case, you'd be out of luck.

Q1: The user would be left to implement tons of boilerplate though, reading the info from stdin, calling the editor, etc. Maybe we should provide a wrapper script, if we do it this way?

Q2: I can't find a way to easily fetch the description of a task. `todo --porcelain show <taskid>` doesn't show it either. It wouldn't break anything if I add the description there, right?

WhyNotHugo:
I won't break anything, adding to `--portecelain` is fine, the only restriction is to avoid deleting or changing stuff there.

PR #314, opened 2018-07-04
Add 'description' and 'modified_at' attributes to the output of --porcelain

hut:
Add 'description' and 'modified_at' attributes to the output of --porcelain.

Btw, is there a reason why we don't simply add all the attributes of Todo class to the output of porcelain?


WhyNotHugo:
Looks good, but some tests need updating since they fail due the the [unexpected] new fields.

Can you update those, please?

Let me know if you've any questions about them, or how to run them locally. Thanks!

hut:
Sure thing, will update them

hut:
Sorry for the delay, this was kinda low prio, but here are the updated test cases. :)

Happy holidays.

reactorcoremeltdown:
Hi everyone!

I'd really like to get this change merged into the upstream. I have patched this up already inside my custom docker container with todoman, mainly because I rely on some automation that uses description field.

Is there anything I could help you with? This is not urgent by any far stretch, and I hugely appreciate your time and effort!

Thanks :)

PR #316, opened 2018-07-17
Default priority

Mome:
I added a ``default_priority`` value to the configuration. It is treated the same way as ``default_due``.

### DIFF
--- a/todoman/cli.py
+++ b/todoman/cli.py
@@ -362,6 +362,10 @@ def new(ctx, summary, list, todo_properties, read_description, interactive):
     if default_due:
         todo.due = todo.created_at + timedelta(hours=default_due)
 
+    default_priority = ctx.config['main']['default_priority']
+    if default_priority:
+        todo.priority = default_priority
+
     for key, value in todo_properties.items():
         if value:
             setattr(todo, key, value)
--- a/todoman/cli.py
+++ b/todoman/cli.py
@@ -363,7 +363,8 @@ def new(ctx, summary, list, todo_properties, read_description, interactive):
         todo.due = todo.created_at + timedelta(hours=default_due)
 
     default_priority = ctx.config['main']['default_priority']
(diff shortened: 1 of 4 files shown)


todoman/cli.py
```diff
@@ -353,6 +353,10 @@ def new(ctx, summary, list, todo_properties, read_description, interactive):
     if default_due:
         todo.due = todo.created_at + timedelta(hours=default_due)
 
+    default_priority = ctx.config['main']['default_priority']
+    if default_priority:
```
WhyNotHugo: This will be false if `default_priority=0`, so the todo will not have a priority set.

This will only work for non-zero priorities (thought according to the spec, unset == 0, that's not the intention there :stuck_out_tongue:  ).

todoman/cli.py
```diff
@@ -353,6 +353,9 @@ def new(ctx, summary, list, todo_properties, read_description, interactive):
     if default_due:
         todo.due = todo.created_at + timedelta(hours=default_due)
 
+    default_priority = ctx.config['main']['default_priority']
+    todo.priority = default_priority
```
WhyNotHugo: I'd rather leave the `if` as `if default_priority is not None`...

todoman/confspec.ini
```diff
@@ -51,3 +51,7 @@ startable = boolean(default=False)
 
 # When running ``todo`` with no commands, run this command.
 default_command = string(default='list')
+
+# The default priority of a task on creation.
+# Highest priority is 1, lowest priority is 10, and 0 means no priority at all.
+default_priority = integer(default=0)
```
WhyNotHugo: ... And have this default to None.

This was if nothing is specified, we don't add a priority (which keeps the current default).

WhyNotHugo:
Rebased this to master, addresses final comments, and added some tests.

PR #318, opened 2018-07-20
Fix coloring with RGBA color strings

Pikrass:
Simple mistake preventing #rrggbbaa strings to be interpreted correctly.


WhyNotHugo:
Makes perfect sense, good catch!

Looks like a test needs updating, since it assumed that a seven-bit value was invalid (and now it kinda is valid).

Pikrass:
Sorry, I didn't have much time to look into that lately. Thanks for doing it!

WhyNotHugo:
No problem! Thanks for the fix!

--
Hugo Osvaldo Barrera

Issue #354, opened 2019-03-28
Suggestion Namespace


927589452:
A very handy feature would be to define namespaces like vdirsyncer does
like

[work]
path= .calendars/work

[personal]
path= .calendars/personal

and returning the calendars with a prefix to help with standard todolist/calendars

work_personal
personal_personal

927589452:
Though a paragraph in FAQ

like

**Problem:** Multiple Accounts using the same name for calendars
**Solution:** Use the [displayname] feature; a file named "displayname" 
                  containing a string will be used as list and human readable name
                  for todoman

WhyNotHugo:
Looks like this is kinda address by using `displayname`, though the documentation around it could be improved. Let's continue discussion in  #355.

PR #355, opened 2019-04-01
Update configure.rst

927589452:
added more on why a displayname should be used

helps with #354


WhyNotHugo:
I feel that it might be good to also link [here](https://vdirsyncer.pimutils.org/en/stable/vdir.html#metadata), which explains this a bit more.

The default is the directory name, but if vdirsyncer is used, it'll be the collection's upstream name.

927589452:
maybe the link to the discussion on why should be replaced by the link you mentioned @WhyNotHugo 
My PR was just a quickfix to stop dumb people like me having dumb problems

Issue #320, opened 2018-08-04
Test fails, if TZ=UTC is not set


dvzrv:
Currently the following test fails, if not prefixed by an `export TZ=UTC`:

```
============================= test session starts ==============================
platform linux -- Python 3.7.0, pytest-3.6.4, py-1.5.4, pluggy-0.7.1
rootdir: /build/todoman/src/todoman-3.4.0, inifile: setup.cfg
plugins: hypothesis-3.66.24
collected 176 items

tests/test_backend.py .........                                          [  5%]
tests/test_cli.py .............x..........F...........................s. [ 35%]
...........                                                              [ 42%]
tests/test_config.py .......xx......                                     [ 50%]
tests/test_filtering.py .........                                        [ 55%]
tests/test_formatter.py ...............                                  [ 64%]
tests/test_main.py .                                                     [ 64%]
tests/test_model.py ............ss..ss................                   [ 84%]
tests/test_porcelain.py ........                                         [ 88%]
tests/test_ui.py ...........                                             [ 94%]
tests/test_widgets.py .........                                          [100%]

=================================== FAILURES ===================================
______________________________ test_default_due2 _______________________________

tmpdir = local('/tmp/pytest-of-builduser/pytest-0/test_default_due20')
runner = <conftest.runner.<locals>.SleepyCliRunner object at 0x6fac5c3d2d30>
create = <function create.<locals>.inner at 0x6fac5c466d08>
todos = {'aaa': <todoman.model.Todo object at 0x6fac5c3be438>, 'bbb': <todoman.model.Todo object at 0x6fac5c3be198>, 'ccc': <todoman.model.Todo object at 0x6fac5c3be400>}

    @pyicu_sensitive
    @freeze_time(datetime.datetime.now())
    def test_default_due2(tmpdir, runner, create, todos):
        cfg = tmpdir.join('config')
        cfg.write('default_due = 24\n', 'a')
    
        r = runner.invoke(cli, ['new', '-ldefault', '-dtomorrow', 'aaa'])
        assert not r.exception
        r = runner.invoke(cli, ['new', '-ldefault', 'bbb'])
        assert not r.exception
        r = runner.invoke(cli, ['new', '-ldefault', '-d', 'one hour', 'ccc'])
        assert not r.exception
    
        todos = {t.summary: t for t in todos(status='ANY')}
        assert todos['aaa'].due.date() == todos['bbb'].due.date()
[...]

PR #359, opened 2019-08-24
Use a defined sourceTime when using parsedatetime

nicoe:
When using the parsedatetime functionalities freezegun is not used to
compute the source time. If the tests are run too close to midnight it
might result in the computation of 'tomorrow' being another day then
now() + 24h (because time.localtime() will be the next day) … thus
failing test_default_due2.

Setting sourceTime when calling parsedatime will ensure the same base is
used for all tests.

### DIFF
--- a/todoman/formatters.py
+++ b/todoman/formatters.py
@@ -192,7 +192,8 @@ class DefaultFormatter:
         except ValueError:
             pass
 
-        rv, pd_ctx = self._parsedatetime_calendar.parse(dt)
+        now = datetime.datetime.now(self.tz)
+        rv, pd_ctx = self._parsedatetime_calendar.parse(dt, now)
         if not pd_ctx.hasDateOrTime:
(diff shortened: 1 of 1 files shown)


nicoe:
AFAICT this should fix #320

WhyNotHugo:
I'd rather decorate affected tests, and keep non-test code as is in this case.

Do you know exactly which tests are affected by this?

nicoe:
* Hugo Osvaldo Barrera [2019-08-26 16:17:23]:
> I'd rather decorate affected tests, and keep non-test code as is in this case.
> 
> Do you know exactly which tests are affected by this?

test_default_due2 and I just noticed that it fails also at this time
and hour when the timezone is CEST but on the second test:

    assert todos['ccc'].due == todos['bbb'].due - datetime.timedelta(hours=23)

(it does not fail with "my" patch though)

PR #360, opened 2019-08-30
Add support for date

### LINKED ISSUE #165: Todoman needs to deal with date vs datetime
Cache stores stuff as timestamps, so when reading back, we always get a `datetime`. We need to find out how to deal with this (maybe a second column stating if it's a dt or not?).

nicoe:
This PR should fix #165

### DIFF
--- a/todoman/formatters.py
+++ b/todoman/formatters.py
@@ -181,6 +181,7 @@ class DefaultFormatter:
             )
 
     def _parse_datetime_naive(self, dt):
+        """Parse dt and returns a naive datetime or a date"""
         try:
             return datetime.datetime.strptime(dt, self.datetime_format)
         except ValueError:
--- a/todoman/formatters.py
+++ b/todoman/formatters.py
@@ -66,7 +66,10 @@ class DefaultFormatter:
             priority = self.format_priority_compact(todo.priority)
 
             due = self.format_datetime(todo.due)
-            if todo.due and todo.due <= self.now and not todo.is_completed:
+            now = (self.now
+                if isinstance(todo.due, datetime.datetime)
+                else self.now.date())
+            if todo.due and todo.due <= now and not todo.is_completed:
                 due = click.style(due, fg='red')
 
             recurring = '⟳' if todo.is_recurring else ''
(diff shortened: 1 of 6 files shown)


todoman/formatters.py
```diff
@@ -180,7 +185,7 @@ def _parse_datetime_naive(self, dt):
             pass
 
         try:
-            return datetime.datetime.strptime(dt, self.date_format)
+            return datetime.datetime.strptime(dt, self.date_format).date()
```
WhyNotHugo: Can you add a docstring to this method? The return value is not as obvious as the name implies any more.

todoman/model.py
```diff
@@ -474,7 +474,9 @@ def create_tables(self):
                 "uid" TEXT,
                 "summary" TEXT,
                 "due" INTEGER,
+                "due_dt" INTEGER,
```
WhyNotHugo: You'll need to push `SCHEMA_VERSION` so that existing installations re-run cache table creation.

todoman/model.py
```diff
@@ -556,14 +558,15 @@ def add_file(self, list_name, path, mtime):
     def _serialize_datetime(self, todo, field):
         dt = todo.decoded(field, None)
         if not dt:
-            return None
+            return None, None
```
WhyNotHugo: Docstring on this function describing the retun values, please?

nicoe:
> ```
> 
> ```
> 
> 
> Can you also add a test reading from an `ical` file which is date-only?

In fact there is already a test for this case in : https://github.com/pimutils/todoman/blob/master/tests/test_model.py#L53 no ?

WhyNotHugo:
Oh, want to add yourself to AUTHORS?

nicoe:
* Hugo Osvaldo Barrera [2019-09-03 11:01:24]:
> Oh, want to add yourself to AUTHORS?

I did both of your suggestions :).
Thanks for help.

nicoe:
Is there anything preventing to merge this PR?

WhyNotHugo:
CI seems to be stuck, for some reason.

I'll trigger it again right now, if it doesn't run properly in an hour I'll just go ahead and merge it.

PR #374, opened 2020-01-20
Fix #301

### LINKED ISSUE #301: Add ability to hide task list name (f.e. when using only one task list)

Phenitei:
Fixes issue #301, as well as some other minor fixes: a few typos, and two flake8 errors.

As a side note, shouldn't the copyright in the docs be updated from 2017 to 2020 ?

Cheers, and thanks for the work,

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -13,21 +13,23 @@ First of all, the classic usage output:
 The default action is ``list``, which outputs all tasks for all calendars, each
 with a semi-permanent unique id::
 
-    1 [ ] !!! 2015-04-30 Close bank account (0%)
-    2 [ ] !				 Send minipimer back for warranty replacement (0%)
-    3 [X] 2015-03-29 Buy soy milk (100%)
-    4 [ ] !!             Fix the iPad's screen (0%)
-    5 [ ] !!             Fix the Touchad battery (0%)
+    1 [ ] !!! 2015-04-30 Close bank account @work (0%)
+    2 [ ] !              Send minipimer back for warranty replacement @home (0%)
+    3 [X]     2015-03-29 Buy soy milk @home (100%)
+    4 [ ] !!             Fix the iPad's screen @home (0%)
+    5 [ ] !!             Fix the Touchpad battery @work (0%)
 
 The columns, in order, are:
 
  * An id.
  * Whether the task has been completed or not.
- * An ``!!!`` indicating high priority,``!!`` indicating medium priority,
+ * An ``!!!`` indicating high priority, ``!!`` indicating medium priority,
    ``!`` indicating low priority tasks.
- * The due date
- * The task summary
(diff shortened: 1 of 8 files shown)


docs/source/usage.rst
```diff
@@ -13,21 +13,23 @@ First of all, the classic usage output:
 The default action is ``list``, which outputs all tasks for all calendars, each
 with a semi-permanent unique id::
 
-    1 [ ] !!! 2015-04-30 Close bank account (0%)
-    2 [ ] !				 Send minipimer back for warranty replacement (0%)
```
WhyNotHugo: Wow, nice catch!

todoman/model.py
```diff
@@ -916,15 +916,15 @@ def colour_for_path(path):
         try:
             with open(os.path.join(path, 'color')) as f:
                 return f.read().strip()
-        except (OSError, IOError):
+        except (IOError):
```
WhyNotHugo: Can you explain this change? It's not obvious to me why `OSError` shouldn't be here.

Also, parens are redudant.

```suggestion
        except IOError:
```

Phenitei: This changes originates in the flake8 report generated by running `tox -e flake8`, which marks this as redundant and suggests using `except ():`, which of course is wrong. After some digging around, as mentioned in the commit, I came across an issue in the flake8 repo explaining this warning: in python2, IOError and OSError are different, but in python3, they actually resolve to the same object:

> Changed in version 3.3: EnvironmentError, IOError, WindowsError, socket.error, select.error and mmap.error have been merged into OSError, and the constructor may return a subclass.

See the discussion in [the issue](https://github.com/PyCQA/flake8-bugbear/issues/110) for details.

Since we only support python3, it makes sense to only use OSError. And of course, because brain, I wrote IOError. What we want is `except OSError:`, I'll fix this.

Phenitei:
Also, is something wrong with the CI ? How come no checks are run ?

todoman/model.py
Phenitei: See 530dc3652445ffb88ee28ed90a4366a411e2ffa2 for the correct change ^^ And applying this PR should also fix the flake8 warning you get in #375.

WhyNotHugo: Ah, thanks for the background on this, good to know!

WhyNotHugo:
Odd. I've no idea what's up with CI not running on your fork. I'll do some digging tomorrow to figure out if forks need to touch anything.

I've copied the branch over [here](https://github.com/pimutils/todoman/runs/402261371) so that CI runs in the meantime.

WhyNotHugo:
Huh, looks like GH realised the commits from my branch were the same as this PR and linked the CI run. :thinking:

WhyNotHugo:
Thanks for all the improvements and fixes! :)

Issue #376, opened 2020-01-23
`todo --porcelain` throws Attributeerror when DUE is a date, not a datetime


benmoran:
I'm using Tasks on Android along with todoman:
https://play.google.com/store/apps/details?id=org.tasks&hl=en_US

It can set due dates as dates with no time.

It seems that the porcelain formatter doesn't like this:

```
  File "/home/ben/.local/pipx/venvs/todoman/lib/python3.7/site-packages/todoman/cli.py", line 645, in list
    click.echo(ctx.formatter.compact_multiple(todos))
  File "/home/ben/.local/pipx/venvs/todoman/lib/python3.7/site-packages/todoman/formatters.py", line 243, in compact_multiple
    data = [self._todo_as_dict(todo) for todo in todos]
  File "/home/ben/.local/pipx/venvs/todoman/lib/python3.7/site-packages/todoman/formatters.py", line 243, in <listcomp>
    data = [self._todo_as_dict(todo) for todo in todos]
  File "/home/ben/.local/pipx/venvs/todoman/lib/python3.7/site-packages/todoman/formatters.py", line 230, in _todo_as_dict
    due=self.format_datetime(todo.due),
  File "/home/ben/.local/pipx/venvs/todoman/lib/python3.7/site-packages/todoman/formatters.py", line 265, in format_datetime
    return int(date.timestamp())
AttributeError: 'datetime.date' object has no attribute 'timestamp'
```

WhyNotHugo:
Fixed by  #377

PR #377, opened 2020-01-23
Issue 376 porcelain date

benmoran:
Fix for #376 by explicitly checking for datetime.date types like the other Formatters do.

(I added myself to AUTHORS.rst in accordance with https://todoman.readthedocs.io/en/stable/contributing.html though it seems a bit much for one line!)

### DIFF
--- a/todoman/formatters.py
+++ b/todoman/formatters.py
@@ -270,7 +270,8 @@ class PorcelainFormatter(DefaultFormatter):
 
     def format_datetime(self, date):
         if date:
-            date = datetime.datetime.fromisoformat(date.isoformat())
+            if not hasattr(date, 'timestamp'):
+                date = datetime.datetime.fromordinal(date.toordinal())
             return int(date.timestamp())
         else:
             return None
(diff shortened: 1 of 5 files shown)


todoman/formatters.py
```diff
@@ -270,6 +270,8 @@ def detailed(self, todo):
 
     def format_datetime(self, date):
         if date:
+            if isinstance(date, datetime.date):
```
WhyNotHugo: This `if` won't do what you expect it.

```
In [3]: isinstance(datetime.now(), date)                                                                                                           
Out[3]: True
```

However, the code works as expected, since the line below does not alter `date` if it's a `datetime`. So it removing the `if` would be cleanest.

benmoran: Good spot, I removed the if.

WhyNotHugo:
Can you merge `master` into your branch? It has a fix so that CI runs properly for PRs.

benmoran:
Couldn't see it on `master` but I did `git merge upstream/run-ci-on-prs`

WhyNotHugo:
Thanks! :)

PR #391, opened 2020-04-28
Improve usage of list comprehensions

WhyNotHugo:


WhyNotHugo:
There's another branch with these same changes.

PR #412, opened 2020-10-19
Update support Pythons to include 3.9

WhyNotHugo:


WhyNotHugo:
Looks like there's no Python3.9 package for macOS yet.

PR #425, opened 2021-03-09
fix bug when default_command has arguments

bernieke:


WhyNotHugo:
Huh, I was sure this same change had already been applied.

Looks perfect, thanks!

PR #436, opened 2021-05-20
pass arguments properly to invoke

bernieke:
The arguments being passed to invoke weren't doing anything.

Now they are properly parsed, converted (sort as list, due as integer, ...) and passed to invoke.

Example from my config file:
```
default_command = "list --sort=due --due 168 --priority low --no-reverse tasks"
```

The arguments to list would be ignored without these changes.


WhyNotHugo:
Looks good. I've added a tests based on your example. Once that runs fine, I'll go ahead and merge this.

Thanks!

WhyNotHugo:
Thanks!

PR #449, opened 2021-08-04
Add option for printing the description in list

gustafla:
I'm opening a pull request about this, not to get my hack without tests and documentation merged, but perhaps to get tips on how to achieve what I want for my todoman workflow. This is pretty much the best I can do with python in an hour, I'm pretty unexperienced at not-so-strongly typed languages. Don't merge this as is :D 

Thanks for creating todoman, it's a great tool to organize my job. Thing is, I want to also see the descriptions of my items in lists to refresh my memory about my "subtasks"/checklists (which OpenTasks on Android supports). I'm thinking the easiest for me to actually implement this feature would be to write a custom wrapper for todoman's --porcelain which could then pretty-print in just the way(s) I want it to.

### DIFF
--- a/todoman/cli.py
+++ b/todoman/cli.py
@@ -607,6 +607,14 @@ def move(ctx, list, ids):
         '"NEEDS-ACTION", "CANCELLED", "COMPLETED", "IN-PROCESS" or "ANY"'
     ),
 )
+@click.option(
+    "--description/--no-description",
+    "-v",
+    default=False,
+    help=(
+        "Show description. Defaults to false."
+    ),
+)
 @catch_errors
 def list(ctx, *args, **kwargs):
     """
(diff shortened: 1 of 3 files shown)


WhyNotHugo:
If you're going to use `--porcelain`, then the description is already present.

I'm not sure about rendering it in the usual view -- the tricky part is making it fit into the layout, especially considering than descriptions can have multiple lines.

Given that we use a table-like layout, maybe putting that in the next row, something like:

```
 5  [ ]    8 days ago   buy breakfast
                        the description shows up here and Lorem ipsum dolor sit
                        amet, consectetur adipiscing elit, sed do eiusmod
                        tempor incididunt ut labore et dolore magna aliqua. Ut
                        enim ad minim veniam, quis nostrud exercitation ullamco
                        laboris nisi ut aliquip ex ea commodo consequat.
 2  [ ]    9 days ago   buy pants
                        This other one fits here and I think you get the idea
```

I don't think that's terribly difficult. But it'd definitely stay behind a flag, since it's too verbose for the default view.

Thoughts?

Issue #297, opened 2018-01-18
Document that list colors need a compatible terminal


untitaker:
In e.g. #251 there has been some confusion about list colors not showing. We could:
* document this (https://todoman.readthedocs.io/en/stable/configure.html#color-and-displayname)
* add a command that will output test colours

or, alternatively:
* round colors to the next 16-bit color for terminals that don't support it. This will loose colour information, not sure if desirable

untitaker:
@WhyNotHugo please advise if you think it's better to support more terminals or to use 256-bit colors. Theoretically we can do both (have 256-bit colors but have a fallback for old terminals) but it'd be messy

WhyNotHugo:
I'd stick to documenting it as a requirement. Not sure if the command is worth it, but I wouldn't be against a `check` command that checks this (I say `check`, because we might want to extend it to other checks that come up in future).

I'm open to PRs for supporting terminals that don't support 24-bit colours though, if somebody cares enough, but I wouldn't worry about it ourselves.

PR #462, opened 2021-10-23
Add support for ansi named colors

Misterio77:
Hey!

I frequently change my terminal colors, and setting RGB colors for my todo folders often leads to unreadable text.

I thought it would be nice to offer an option for setting one of the usual 16 terminal colors, in addition to the 24-bit rgb ones. So, i've added support for them.

It works for all the 16 ansi colors (including the 8 bright ones).

This possibly fixes https://github.com/pimutils/todoman/issues/297

Here's how it looks like (using dracula scheme, and colors `red`, `cyan`, and `bright purple` (which my terminal renders as bold purple) for the folders):
image

### DIFF
--- a/todoman/formatters.py
+++ b/todoman/formatters.py
@@ -19,7 +19,7 @@ from todoman.model import Todo
 from todoman.model import TodoList
 
 
-def rgb_to_ansi(colour: str) -> Optional[str]:
+def rgb_to_ansi(colour: str) -> str:
     """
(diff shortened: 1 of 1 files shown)


Misterio77:
Thanks for the feedback.

I wasn't aware the color setting was something standardized. I guess my provider doesn't support syncing colors.

Khal has a [config option for manually setting colors](https://lostpackets.de/khal/configure.html), so I guess something similar could be implemented here, would that make sense?

PR #468, opened 2021-11-29
When we read in binary mode, write using it as well, fixes writing under Windows

tyrylu:
Under Windows, the distinction between binary or text mode actually exists, this PR allows to run the creation commands under Windows.

### DIFF
--- a/todoman/model.py
+++ b/todoman/model.py
@@ -390,8 +390,8 @@ class VtodoWriter:
                 if component.get("uid", None) == self.todo.uid:
                     cal.subcomponents[index] = vtodo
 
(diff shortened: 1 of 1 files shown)


WhyNotHugo:
This makes sense; it also avoids a decode+encode cycle that's not necessary.

I guess files still end up UTF-8 encoded, right?

WhyNotHugo:
Thanks!

tyrylu:
They should, but it depends on the implementation in vobject. However, the decode was depending on that behavior as well, even if in a slightly different way, so, yes, the files will end up as UTF-8.

PR #482, opened 2022-04-06
Added "read from stdin" option to edit command

reactorcoremeltdown:
Added `-r` (`--read-description`) flag to `edit` command.

This option is vital for my workflow: I use todoman inside a container to sync data between Calendars, Tasks, and also Kanboard. I need to sync back kanboard cards' URLs to tasks' descriptions.


WhyNotHugo:
Applied via https://github.com/pimutils/todoman/pull/587

PR #495, opened 2022-08-28
update zsh completion

balejk:
This is not meant to be comprehensive, but I fixed several issues that I
encountered when trying to use the completion. Most notably, I have adapted it
to work with the new Python configuration file format.


WhyNotHugo:
Thanks

PR #512, opened 2023-01-17
Switch to ruff for linting

WhyNotHugo:
Ruff is a newer linter which re-implements flake8 (plus a few of its plugins),
isort and pyupgrade into one single codebase. It's stupid fast, and after some
testing is proving to be quite reliable.


WhyNotHugo:
Merged via 6c9f8d4a0534c8b3a1edc8def2f6ee979970befb

Issue #485, opened 2022-04-21
meta: Recurring events


WhyNotHugo:
Meta issue to track all other issues regarding recurring events:

- [ ] https://github.com/pimutils/todoman/issues/305
- [ ] https://github.com/pimutils/todoman/issues/410
- [ ] https://github.com/pimutils/todoman/issues/305
- [ ] https://github.com/pimutils/todoman/issues/304
- [ ] https://github.com/pimutils/todoman/issues/288
- [ ] https://github.com/pimutils/todoman/issues/249

Issue #288, opened 2017-10-07
Set "RELATED-TO" for completed recurring todos


WhyNotHugo:
When a recurring todo is completed, we create a copy of it. The old copy is marked as done, and the new copy keeps the `RRULE` and it set for the next recurrance.

However, nothing really relates these events. Let's use the [`RELATED-TO`](https://tools.ietf.org/html/rfc5545#section-3.8.4.5) attribute, which makes perfect sense for this.

ngortheone:
I think `RELATED-DO` might not be the best option for this. There is a conflict with
https://github.com/pimutils/todoman/issues/396

For example, tasks that I migrated from Akonadi (KDE) use `RELATED-TO` to establish task hierarchy

WhyNotHugo:
This was implemented in 2017. From [rfc5545#section-3.8.4.5](https://datatracker.ietf.org/doc/html/rfc5545#section-3.8.4.5):

> By default, the property value points to another calendar
> component that has a PARENT relationship to the referencing
> object.  The "RELTYPE" property parameter is used to either
> explicitly state the default PARENT relationship type to the
> referenced calendar component or to override the default PARENT
> relationship type and specify either a CHILD or SIBLING
> relationship.  The PARENT relationship indicates that the calendar
> component is a subordinate of the referenced calendar component.
> The CHILD relationship indicates that the calendar component is a
> superior of the referenced calendar component.  The SIBLING
> relationship indicates that the calendar component is a peer of
> the referenced calendar component.

To avoid confusion, I think it would be sane to add `RELTYPE`for recurring relationships, specifically, `RELATED-TO;RELTYPE=SIBLING:XYZ`, where XZY is the UID of the other recurring instance... though maybe an `X-` attribute would make sense?

I'll try and consult what other task applications do and try to be consistent.

@ngortheone Can you confirm that Akonadi sets the RELTYPE=PARENT`? If not, it definitely should.

PR #520, opened 2023-04-07
Implement undo

r4ulill0:
The undo functionality does not really resolve recurring todos because that functionality is not implemented yet so there is a test that will fail until that is implemented (I'm planing to solve that myself, but one thing at a time).

@WhyNotHugo tell me if you see something off and sorry for taking that long, recurring todos distracted me a bit :laughing:


todoman/model.py
```diff
@@ -252,6 +252,27 @@ def complete(self) -> None:
         self.percent_complete = 100
         self.status = "COMPLETED"
 
+    def undo(self) -> Todo | None:
+        """
+        Immediately restores this todo. Marks it as needs action, resets the
+        percentage to 0 and deletes the completed_at datetime.
+
```
WhyNotHugo, line 259:
```suggestion
        """Immediately restores this todo.

        Marks it as needs action, resets the percentage to 0 and deletes the
        completed_at datetime.
```

The first line is show in short help output.

WhyNotHugo (approved):

r4ulill0:
> I don't fully understand why one tests is failing: https://builds.sr.ht/~whynothugo/job/981248#task-test-262

Recurring test doesn't work because there are some recursion fields that are not persisted yet. We have to deal some issues  #485 in order for it to work. I'm planning on taking all of them but one step at a time.

I tried to implement persistence for RELATED-TO but it was a pretty big change and it has its own issue #288.
You can open a new issue to fix this test and add it to #485 if you want. Also we can delete the test until its relevant. I'm fine with either option.

r4ulill0:
I have marked the test as the RELATED-TO property needs to be persisted before it can work.

Now arch build passes (Alpine build just exploded and I don't understand why).

dikey0ficial:
Is there any progress?..

r4ulill0:
> Is there any progress?..

It should be ready to merge. I don't know if @WhyNotHugo has any more things to check.

todoman/cli.py
```diff
@@ -469,6 +469,30 @@ def done(ctx, todos):
...
+@cli.command()
+@pass_ctx
+@click.argument(
+    "todos",
+    nargs=-1,
+    required=True,
+    type=click.IntRange(0),
+    callback=_validate_todos,
+)
+# @catch_errors
```
WhyNotHugo, line 481: Is this bit problematic?

todoman/model.py
```diff
@@ -250,6 +250,27 @@ def complete(self) -> None:
...
 
+    def undo(self) -> Todo | None:
+        """Immediately restores this todo.
+
+        Marks it as needs action, resets the percentage to 0 and deletes the
+        completed_at datetime.
+        If this todo belongs to a series, the one created at completion will
+        be deleted.
+
+        Returns the todo that will be deleted.
```
WhyNotHugo, line 261:
```suggestion
        Returns the todo that should be deleted.
```

todoman/model.py
```diff
@@ -250,6 +250,27 @@ def complete(self) -> None:
...
+        """
+        original = None
+        if self.is_recurring and self.related:
+            original = self.related.pop()
+            self.rrule = original.rrule
```
WhyNotHugo, line 266: Isn't `original` the newer entry here?

tests/test_cli.py
```diff
@@ -747,6 +748,55 @@ def test_done_recurring(runner, todo_factory, todos):
...
+    result = runner.invoke(cli, ["undo", "1"])
+    assert not result.exception
+
+    todo = next(todos(status="ANY"))
+    assert todo.percent_complete == 0
+    assert todo.is_completed is False
+    assert not todo.rrule
+
+
+@pytest.mark.xfail(reason="Not implemented")
```
WhyNotHugo, line 765: I think that recursion is now implemented, right?

r4ulill0: No, it is not.

There are multiple PRs regarding recursion. The most relevant here is #288 but there are some more listed in #485.

todoman/cli.py
r4ulill0: It doesn't. Restored

todoman/model.py
r4ulill0: Yeah, bad naming. Solved.

todoman/model.py
```diff
@@ -250,6 +250,27 @@ def complete(self) -> None:
...
+
+        Marks it as needs action, resets the percentage to 0 and deletes the
+        completed_at datetime.
+        If this todo belongs to a series, the one created at completion will
+        be deleted.
+
+        Returns the todo that should be deleted.
+        """
+        recurred = None
+        if self.is_recurring and self.related:
```
WhyNotHugo, line 264: I ran:

```sh
todo new -l test test
todo new 1
```

The completed todo looks like this:

```icalendar
BEGIN:VCALENDAR
VERSION:2.0
PRODID:io.barrera.todoman
BEGIN:VTODO
COMPLETED:20231118T090119Z
CREATED:20231118T090107Z
DTSTAMP:20231118T090107Z
DUE:20231119T090107Z
LAST-MODIFIED:20231118T090119Z
PERCENT-COMPLETE:100
SEQUENCE:2
STATUS:COMPLETED
SUMMARY:test
UID:1d54c495739f4546a92974ab574b5347@destiny.whynothugo.nl
END:VTODO
END:VCALENDAR
```

It doesn't have a recurrence rule any more (that is moved to the next one in the series), so this condition does not match.

If I run `todo undo 1`, both todos are kept around.

This is not ideal, but I don't think that it is terrible either. The RFC doesn't even cover how a recurring TODO is completed, so there's no correct behavior to any of this either way.

However, this `if` will never run, so this code is misleading. I think it should be removed (unless I'm missing something here).

PR #521, opened 2023-04-08
Mst/zsh completion

mstmob:
Update contrib/completion/zsh
- Add existing categories completion
- Unify options in common options
- Fix typos


WhyNotHugo:
Tested this a bit and seems to work fine. Thanks!

PR #533, opened 2023-07-08
Fix CI failing on Arch

WhyNotHugo:
`pip` can't be used any more on Arch, so use `pipx` instead.


WhyNotHugo:
Merged via 8317c05684bcaba2bc53c0e1c41e5d2bc47bcf70

PR #541, opened 2023-09-06
fix filtering for non-ASCII categories

### LINKED ISSUE #540: categories with diacritics not handled well
Steps to reproduce:
1. Create a new task via interactive mode and set its category to `ý`.
2. Run `todo list -c ý`
3. The task is not listed.

I have tried setting `🤷` as a category as well to check whether it could be some more general encoding error, but this works.

I am using version 4.3.1 from the Void Linux repository.

balejk:
To make category filtering case insensitive, todoman converts the
category provided on the commandline to uppercase, uses the sqlite
`upper` function to convert the cache entries and then compares the
converted strings.

However the sqlite `upper` function only works for ASCII strings causing
problems when the category is for instance `ý`, which sqlite translates
to uppercase as `ý` while the correct version is `Ý`, causing the
filtering for this category not to return any tasks.

Add a new column (`category_upper`) to the `categories` table with the
uppercase version of the category name converted by Python (which
performs the conversion correctly) and use this for filtering while
keeping the original version in the `category` column as before.

Add a corresponding test.

---
Fixes #540.

### DIFF
--- a/tests/test_filtering.py
+++ b/tests/test_filtering.py
@@ -72,11 +72,20 @@ def test_category(tmpdir, runner, create):
     create("one.ics", f"UID:{uuid4()}\nSUMMARY:haha\nCATEGORIES:work,trip\n")
     create("two.ics", "CATEGORIES:trip\nSUMMARY:hoho\n")
     create("three.ics", f"UID:{uuid4()}\nSUMMARY:harhar\n")
+    create("four.ics", f"UID:{uuid4()}\nSUMMARY:harharho\nCATEGORIES:ý,Дом\n")
     result = runner.invoke(cli, ["list", "--category", "work"])
     assert not result.exception
     assert "haha" in result.output
     assert "hoho" not in result.output
     assert "harhar" not in result.output
+
+    # non-ASCII categories
+    result = runner.invoke(cli, ["list", "--category", "ý"])
+    assert not result.exception
+    assert "harharho" in result.output
+    result = runner.invoke(cli, ["list", "--category", "Дом"])
+    assert not result.exception
+    assert "harharho" in result.output
 
 
 def test_grep(tmpdir, runner, create):
--- a/todoman/model.py
+++ b/todoman/model.py
@@ -402,7 +402,7 @@ class Cache:
     may be used for filtering/sorting.
     """
 
-    SCHEMA_VERSION = 9
+    SCHEMA_VERSION = 10
 
     def __init__(self, path: str):
         self.cache_path = str(path)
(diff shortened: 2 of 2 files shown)


balejk:
Is there a way to force cache regeneration? As it is now, this causes "column does not exist" error after update.

WhyNotHugo:
Just delete the cache and it'll be regenerated.

If you bump[ `Cache.SCHEMA_VERSION`](https://github.com/pimutils/todoman/blob/93af8a67faaffa1709ee5467b45a7a5e0bce640d/todoman/model.py#L405), todoman will recognise all caches as old and regenerate them automatically. You should increase the `SCHEMA_VERSION` to `10` for this PR so the cache is regenerated seamlessly for users when upgrading.

tests/test_filtering.py
```diff
@@ -72,12 +72,21 @@ def test_category(tmpdir, runner, create):
     create("one.ics", f"UID:{uuid4()}\nSUMMARY:haha\nCATEGORIES:work,trip\n")
     create("two.ics", "CATEGORIES:trip\nSUMMARY:hoho\n")
     create("three.ics", f"UID:{uuid4()}\nSUMMARY:harhar\n")
+    create("four.ics", f"UID:{uuid4()}\nSUMMARY:harharho\nCATEGORIES:ý,Дом\n")
```
WhyNotHugo, line 75: Just to be sure: `Д` is uppercase, right? I definitely want to keep an uppercase Cyrillic letter in tests.

WhyNotHugo (approved):

balejk:
> Just delete the cache and it'll be regenerated.

Yes, of course, I meant for production.

> If you bump[ `Cache.SCHEMA_VERSION`](https://github.com/pimutils/todoman/blob/93af8a67faaffa1709ee5467b45a7a5e0bce640d/todoman/model.py#L405), todoman will recognise all caches as old and regenerate them automatically. You should increase the `SCHEMA_VERSION` to `10` for this PR so the cache is regenerated seamlessly for users when upgrading.

I will push this in a second.

balejk:
> Just to be sure: `Д` is uppercase, right? I definitely want to keep an uppercase Cyrillic letter in tests.

But actually, I think you rather want to make sure there is such a
lowercase letter, no?

WhyNotHugo:
> But actually, I think you rather want to make sure there is such a lowercase letter, no?

I wanted to make sure that we have both lowercase AND uppercase in tests.

WhyNotHugo:
@balejk You didn't add yourself to `AUTHORS.rst`.

balejk:
@WhyNotHugo I see, sorry. Feel free to add me if it's required (or I can open another PR, but that seems needlessly complicated).

PR #549, opened 2024-02-29
Change formatting to two pass so that columns can be aligned and provide text wrapping to tty column size.

elric1:


WhyNotHugo:
We used to have table rendering, but it was removed in 586c0c2edb01893b56316ab4de0356c694a179dd.

Maybe we can bring it back behind a config option (e.g.: make it optional). I think the `Formatter` plumbing might work for this.

elric1:
Oh, I didn't see that. It does end up being a little different than what I was doing, in a number of ways:
1. the [X] for completed was flipped in 586c0c2 and I preserved that not knowing that it was flipped.
2. I'm wrapping the summary/categories column. Tabulate can't wrap in the way that I'm doing as it requires you to specify a maxcolwidth for each column, but can't do it for the whole screen. This is spoken of in [Tablulate Issue #290](https://github.com/astanin/python-tabulate/issues/290), but the proposed solution there is also suboptimal as it would merely set the maxcolwidth to tty cols / num cols which would leave the summary column way too small.
3. I'm entirely leaving out columns that don't exist which makes it a bit more compact. I could still do this with tabulate, though, but it would still be a two pass algorithm.

Is using tabulate important for this patch, or is it acceptable to just format it ourselves? Tabulate appears to provide quite a few extra types of output which could be quite nice to offer such as HTML or LaTeX, so maybe the best strategy would be to extend tabulate and submit a PR with them and use it?

elric1:
In accordance with your comments above, I have changed approach. Now, I submit a patch to use tabulate, but I make it configurable so that the TableFormat can be selected by the user. They can therefore use plain, html, markdown, etc. I also submitted a pull request to the tabulate project to add a "flow" table type which does not align the columns and drops empty columns: [Tabulate PR#314](https://github.com/astanin/python-tabulate/pull/314)

WhyNotHugo:
Tests indicate that there are extra spaces in the default outputs: https://builds.sr.ht/~whynothugo/job/1161850#task-test-517

WhyNotHugo:
You can run tests locally with `tox -e py`.

WhyNotHugo:
Feel free to ask if you have issues running tests.

WhyNotHugo:
This approach is good, using `tabulate` is a lot less code and I know it works well already :+1:

mgajda:
Can also add a function to discover TTY size (on Linux: either COLUMNS variable or PTY). I can provide patch for this.

PR #560, opened 2024-08-10
Remove deprecated pytest-runner

WhyNotHugo:
We don't seem to actually be relying on this anyway.

Fixes: https://github.com/pimutils/todoman/issues/559


WhyNotHugo:
Merged via https://github.com/pimutils/todoman/commit/e90cd83993e2973db19abd56e0f529aeb6ec9e2a

PR #563, opened 2024-10-01
explicitly depend on pytz

dotlambda:
icalendar removed the dependency in v6


WhyNotHugo:
Fixed by dropping pytz. Thanks for bringing this to my attention.

PR #564, opened 2024-10-10
fix(pyproject): add pytz dep

chenrui333:
missing pytz dep in pyproject

relates to https://github.com/Homebrew/homebrew-core/pull/193639


chenrui333:
actually duplicate of #563

PR #581, opened 2025-07-07
Porcelain (JSON) output should include whether an event recurs

dme:


dme:
> Porcelain output is JSON, not YAML. Commit message needs to be ammended.

Yikes, brain fade on my part, apologies. Will reword and resubmit.

Issue #582, opened 2025-07-09
atomicwrites is depricated


AndreasBilke:
atomicwrites is abandoned/deprecated since some time. This resulted in kicked out package in different distros (like FreeBSD).

Other PIM software also made/preparing the changes for replacing that software. See https://github.com/lucc/khard/issues/342 and https://github.com/pimutils/vdirsyncer/pull/1162

Would it be possible to have this package also removed from todoman?

mathstuf:
I'll take this one too while I'm doing this.

PR #583, opened 2025-07-10
model: remove `atomicwrites` dependency

mathstuf:
Lifted from pimutils/khal#1393 (which I also authored).

Fixes: #582


mathstuf:
I tried fixing CI in #584 but I'm at the end of the time I can spend on that.

PR #584, opened 2025-07-10
builds: use alpine/3.19 for Python 3.11

mathstuf:
Alpine 3.18 is EOL.


mathstuf:
Well, that's about the extent of my investigation here…

WhyNotHugo:
Thanks. Applied the first two commits. Fixed the other test.

PR #588, opened 2025-07-20
Docs: `:show-nested:` is deprecated

imomaliev:
From the `sphinx-click` docs

> This option is deprecated; use nested instead.

Read more https://sphinx-click.readthedocs.io/en/latest/usage/#usage


WhyNotHugo:
Applied, thanks

PR #594, opened 2025-09-08
New support for subtasks

### LINKED ISSUE #396: Show subtasks in a better way
It would be great if subtasks would be grouped within their task, e.g. by intending the description or something.

subtasks = use `RELATED-TO`

powerjungle:
Add support for RELATED-TO property and RELTYPE parameter.

Taken from [RFC 5545, section `3.8.4.5`](https://www.rfc-editor.org/rfc/rfc5545#section-3.8.4.5). If present in a file, these values are now read and stored in the cache.

Tests are modified to reflect the change.

The `SCHEMA_VERSION` is incremented to allow the cache to be recreated after the update.

Nothing is required from the user to be done for the changes in this PR to work.

Before:
image

After:
image

Closes: https://github.com/pimutils/todoman/issues/396

### DIFF
--- a/docs/source/contributing.rst
+++ b/docs/source/contributing.rst
@@ -23,6 +23,11 @@ without them being properly documented.
 Run ``pip install -e .`` to install todoman and its dependencies into a
 virtualenv.
 
+If the database is changed in a breaking way, the ``SCHEMA_VERSION`` variable
+in the class ``Cache`` has to be incremented to allow the cache to be recreated
+after todoman has been updated. An example would be adding new fields or
+removing old unnecessary fields and etc.
+
 We use ``pre-commit`` to run style and convention checks. Run ``pre-commit
 install``` to install our git-hooks. These will check code style and inform you
 of any issues when attempting to commit. This will also run ``black`` to
--- a/tests/test_porcelain.py
+++ b/tests/test_porcelain.py
@@ -22,7 +22,8 @@ def test_list_all(tmpdir: py.path.local, runner: CliRunner, create: Callable) ->
         "DUE;VALUE=DATE-TIME;TZID=CET:20160102T000000\n"
         "DTSTART:20160101T000000Z\n"
         "PERCENT-COMPLETE:26\n"
-        "LOCATION:Wherever\n",
+        "LOCATION:Wherever\n"
+        "RELATED-TO;RELTYPE=PARENT:123456789\n",
     )
     result = runner.invoke(cli, ["--porcelain", "list", "--status", "ANY"])
 
@@ -40,6 +41,8 @@ def test_list_all(tmpdir: py.path.local, runner: CliRunner, create: Callable) ->
             "percent": 26,
             "priority": 0,
             "recurring": False,
+            "related_to": "123456789",
+            "related_to_reltype": "PARENT",
             "start": 1451606400,
             "summary": "Do stuff",
         }
@@ -78,6 +81,8 @@ def test_list_start_date(
             "percent": 26,
             "priority": 0,
             "recurring": False,
+            "related_to": "",
+            "related_to_reltype": "",
             "start": 1451692800,
             "summary": "Do stuff",
         }
@@ -114,6 +119,8 @@ def test_list_due_date(
             "percent": 26,
             "priority": 0,
             "recurring": False,
+            "related_to": "",
+            "related_to_reltype": "",
             "start": None,
             "summary": "Do stuff",
         }
@@ -143,6 +150,8 @@ def test_list_nodue(tmpdir: py.path.local, runner: CliRunner, create: Callable) 
             "location": "",
             "percent": 12,
             "recurring": False,
+            "related_to": "",
+            "related_to_reltype": "",
             "priority": 4,
             "start": None,
             "summary": "Do stuff",
@@ -216,6 +225,8 @@ def test_show(tmpdir: py.path.local, runner: CliRunner, create: Callable) -> Non
         "percent": 0,
         "priority": 5,
         "recurring": False,
+        "related_to": "",
+        "related_to_reltype": "",
         "start": None,
         "summary": "harhar",
     }
@@ -241,6 +252,8 @@ def test_simple_action(todo_factory: Callable) -> None:
         "percent": 0,
         "priority": 0,
         "recurring": False,
+        "related_to": "",
+        "related_to_reltype": "",
         "start": None,
         "summary": "YARR!",
     }
--- a/todoman/cli.py
+++ b/todoman/cli.py
@@ -238,6 +238,13 @@ def _todo_property_options(command: Callable) -> Callable:
         help="When the task starts.",
     )(command)
 
+    # Merges all different property arguments into one dictionary
+    # `todo_properties` argument, so that it can be directly looped through
+    # easily for directly setting the `todoman.model.Todo` class attributes
+    # from within a command function.
+    #
+    # The names of the options are the same as the
+    # `todoman.model.Todo` class attributes.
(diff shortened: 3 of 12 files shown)


balejk:
If by cache you mean the cache database, it should be enough to just increment the [schema version](https://github.com/pimutils/todoman/blob/bd417730a6ee761734d07686511c24a1ba7cebda/todoman/model.py#L425) to force a rebuild.

jlkDE:
Hey, I just rebased this to the current main for myself and as far as I can see everything works as intended, thanks for that!
What is needed for this to progress, is there some way to support this?

tests/test_porcelain.py
```diff
@@ -78,6 +81,8 @@ def test_list_start_date(
             "percent": 26,
             "priority": 0,
             "recurring": False,
+            "related_to": "",
+            "related_to_reltype": "",
```
WhyNotHugo, line 85: We should exclude the fields when absent, instead of including empty fields.

(same applies to other instances below)

docs/source/contributing.rst
```diff
@@ -23,6 +23,11 @@ without them being properly documented.
 Run ``pip install -e .`` to install todoman and its dependencies into a
 virtualenv.
 
+If the database is changed in a breaking way, the ``SCHEMA_VERSION`` variable
+in the class ``Cache`` has to be incremented to allow the cache to be recreated
+after todoman has been updated. An example would be adding new fields or
+removing old unnecessary fields and etc.
+
```
WhyNotHugo, line 30: Nice addition!

todoman/cli.py
```diff
@@ -238,6 +240,13 @@ def _todo_property_options(command: Callable) -> Callable:
...
 
+    # Merges all different property arguments into one dictionary
+    # `todo_properties` argument, so that it can be directly looped through
+    # easily for directly setting the `todoman.model.Todo` class attributes
+    # from within a command function.
+    #
+    # The names of the options are the same as the
+    # `todoman.model.Todo` class attributes.
     @functools.wraps(command)
     def command_wrap(*a, **kw) -> click.Command:
```
WhyNotHugo, line 251: Nitpick: this should be a docstring rather than comment.

todoman/cli.py
```diff
@@ -284,6 +293,21 @@ def formatter(self) -> formatters.Formatter:
     help="Go into interactive mode before saving the task.",
 )
 
+_subtask_option = click.option(
+    "--subtask-for",
```
WhyNotHugo, line 297: What do you think of calling this `--parent`?

Code-wise it's a bit more obvious, but I'm not sure how obvious it is for a user interface.

todoman/cli.py
```diff
@@ -515,6 +550,17 @@ def edit(
         changes = True
         todo.description = sys.stdin.read()
 
+    if subtask_for is not None and not_subtask is False:
```
WhyNotHugo, line 554: If `subtask_for and not_subtask`, we should return an error (invalid arguments).

todoman/formatters.py
```diff
@@ -120,51 +120,155 @@ def compact(self, todo: Todo) -> str:
...
+        # or grey (future)
+
+        # Holds information needed to properly order the text output.
+        #
+        # key: UID of todo
+        # value: list which contains 2 elements:
+        #            0: the formatted todo text line for the output
+        #            1: a dictionary which represents child todos which have
+        #               the same structure as the parent
+        tree: dict[str, list] = {}
```
WhyNotHugo, line 133: A tuple is best if items are heterogeneous and a fixed amount:

```suggestion
        tree: dict[str, tuple[str, dict[…]]] = {}
```

todoman/formatters.py
```diff
@@ -120,51 +120,155 @@ def compact(self, todo: Todo) -> str:
...
+
+        # Holds all of the todo relationships in the form of key-value pairs.
+        #
+        # key: child
+        # value: list which contains 2 elements:
+        #            0: parent
+        #            1: any special information like "SIBLING"
+        related_todos: dict[str, list] = {}
```
WhyNotHugo, line 141: Same as above.

todoman/formatters.py
```diff
@@ -120,51 +120,155 @@ def compact(self, todo: Todo) -> str:
...
+            # If RELTYPE is empty, the default is PARENT.
+            # Source:
+            # https://www.rfc-editor.org/rfc/rfc5545#section-3.2.15
+            #
+            # "To preserve backwards compatibility, the value type MUST be
+            # UID when the PARENT, SIBLING, or CHILD relationships
+            # are specified."
+            # Source:
+            # https://www.rfc-editor.org/rfc/rfc9253#section-9.1
+            if todo.related_to != "":
```
WhyNotHugo, line 153: Check truthiness, so this also matches if it's `None`.

```suggestion
            if todo.related_to:
```

todoman/formatters.py
```diff
@@ -120,51 +120,155 @@ def compact(self, todo: Todo) -> str:
...
-                f"[{completed}] {todo.id} {priority} {due} "
-                f"{recurring}{summary}{categories}"
-            )
+        self._tree_reorder_related(tree, related_todos)
 
-        return "\n".join(table)
+        return self._join_tree(tree).strip()
+
+    def _tree_reorder_related(
+        self, tree: dict[str, list], related_todos: dict[str, list]
```
WhyNotHugo, line 205: Formatting:

```suggestion
        self,
        tree: dict[str, list],
        related_todos: dict[str, list],
```

todoman/formatters.py
```diff
@@ -120,51 +120,155 @@ def compact(self, todo: Todo) -> str:
...
+            for inwards in store_path:
+                try:
+                    related_dict = related_dict[inwards][1]
+                    if related_dict is None:
+                        break
+                except KeyError:
+                    related_dict = tree
+                    break
+
+            self._tree_move_related(related, related_to[0], tree, related_dict)
```
WhyNotHugo, line 236: This looks quite inefficient, since it walks the entire list multiple times. I think we need a "real" tree when building the whole tree in the first place.

But this is non-blocking at this stage; we can merge it in its current form and iterate on this later.

todoman/model.py
```diff
@@ -93,6 +95,8 @@ def __init__(
         self.location = ""
         self.percent_complete = 0
         self.priority = 0
+        self.related_to = ""
+        self.related_to_reltype = ""
```
WhyNotHugo, line 99: Default should be `None`, not empty string. This affects the field's presence in the output icalendar file.

todoman/model.py
```diff
@@ -659,6 +681,11 @@ def add_vtodo(
         :param todo: The icalendar component object on which
         """
 
+        got_related_to_reltype = ""
+        got_related_to = todo.get("related-to", "")
+        if got_related_to != "":
+            got_related_to_reltype = got_related_to.params.get("reltype")
+
```
WhyNotHugo, line 693: As above, default should be `None`.

WhyNotHugo (changes requested):
Rebased this applying `ruff format` on the way.

Looks good, appreciate all the attention to detail.

Mostly minor feedback. There are two things which I think can be improved but can be done separate later (I'd rather keep this PR lean and not block on further changes):

- Tree-building. This seems rather inefficient right now, but I won't block merging on that; we can iterate later gradually.
- Style. I think the output would be easier to parse keeping the left `[ ] $ID` fields all aligned, maybe even something like:

```
[ ] 2 Some task
[ ] 4 ├─ Some subtask
[ ] 3 └─ Another subtask
```

Again, this seems like something that's best done separately.

PR #598, opened 2025-12-22
Recurrent items should not acquire a time unnecessarily

dme:
When generating the next instance of a recurring item, avoiding adding a time (00:00:00) to the new item if the originating item did not include one.

### DIFF
--- a/todoman/model.py
+++ b/todoman/model.py
@@ -210,17 +210,22 @@ class Todo:
     def is_recurring(self) -> bool:
         return bool(self.rrule)
 
-    def _apply_recurrence_to_dt(self, dt: date | None) -> datetime | None:
+    def _apply_recurrence_to_dt(self, dt: date | None) -> date | datetime | None:
         if not dt:
             return None
 
         assert self.rrule is not None, "applying recurrence to todo without rrule"
         recurrence = rrulestr(self.rrule, dtstart=dt)
 
-        if isinstance(dt, date) and not isinstance(dt, datetime):
+        if dateonly := not isinstance(dt, datetime):
             dt = datetime.combine(dt, time.min)
 
-        return recurrence.after(dt)
+        nxt = recurrence.after(dt)
+
+        if dateonly and nxt is not None:
+            return nxt.date()
+
+        return nxt
 
     def _create_next_instance(self) -> Todo:
         copy = self.clone()


WhyNotHugo:
[Mypy fails](https://builds.sr.ht/~whynothugo/job/1732439#task-lint-15). Pretty sure it's a false positive, but it can't infer that you're converting `dt` into the correct type:

```
> mypy .
todoman/model.py:224: error: Argument 1 to "after" of "rrulebase" has incompatible type "date"; expected "datetime"  [arg-type]
todoman/model.py:227: error: Item "None" of "datetime | None" has no attribute "date"  [union-attr]
todoman/model.py:227: error: Incompatible types in assignment (expression has type "date | Any", variable has type "datetime | None")  [assignment]
Found 3 errors in 1 file (checked 24 source files)
```

todoman/model.py
```diff
@@ -210,17 +210,23 @@ def is_completed(self) -> bool:
...
-        if isinstance(dt, date) and not isinstance(dt, datetime):
+        dateonly = isinstance(dt, date) and not isinstance(dt, datetime)
+        if dateonly:
             dt = datetime.combine(dt, time.min)
 
-        return recurrence.after(dt)
+        nxt = recurrence.after(dt)
+
+        if dateonly:
+            nxt = nxt.date()
```
WhyNotHugo, line 228: `after()` can return `None` (when there are no more repetitions), so `nxt.date()` crashes.

WhyNotHugo (changes requested):

todoman/model.py
dme: 👍

todoman/model.py
```diff
@@ -210,17 +210,23 @@ def is_completed(self) -> bool:
...
         assert self.rrule is not None, "applying recurrence to todo without rrule"
         recurrence = rrulestr(self.rrule, dtstart=dt)
 
-        if isinstance(dt, date) and not isinstance(dt, datetime):
+        dateonly = isinstance(dt, date) and not isinstance(dt, datetime)
+        if dateonly:
             dt = datetime.combine(dt, time.min)
 
-        return recurrence.after(dt)
+        nxt = recurrence.after(dt)
```
WhyNotHugo, line 225: Let's make mypy shut up. This should always be true.

```suggestion

        assert isinstance(dt, datetime)
        nxt = recurrence.after(dt)
```

WhyNotHugo:
Just that annoying typing issue, otherwise looks good.

edit: the fact that mypy complains is annoying, not your code :)

todoman/model.py
dme: 👍🏻

WhyNotHugo:
Tweaked a little to work around the annoying typing issue.

Thanks!

PR #603, opened 2026-04-14
Update tested environments in CI

WhyNotHugo:


WhyNotHugo:
Tests leak file descriptors (which is not really an issue for production code), and eventually reach a point where CI runs out of FDs.

PR #606, opened 2026-04-30
fix: correct misleading return type annotation on format_datetime

AmSach:
Fixed misleading type annotation in format_datetime method. The abstract method and DefaultFormatter were annotated as returning str | int | None, but int was never actually returned. Only str or None is returned in practice. This prevents type checkers from flagging actual bugs where int is expected but str is received.

### DIFF
--- a/todoman/formatters.py
+++ b/todoman/formatters.py
@@ -67,7 +67,7 @@ class Formatter(ABC):
         """Returns a detailed representation of a task."""
 
     @abstractmethod
-    def format_datetime(self, value: date | None) -> str | int | None:
+    def format_datetime(self, value: date | None) -> str | None:
         """Format an optional datetime."""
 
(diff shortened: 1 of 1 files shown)


WhyNotHugo:
mypy fails: https://builds.sr.ht/~whynothugo/job/1745773

Subclass `PorcelainFormatter` does return `int`.

Arguably, `PorcelainFormatter` should not be a subclass of the regular formatter; it really does have a different API.

WhyNotHugo:
We probably want to have a `class InteractiveFormatter(Formatter)` which declares these methods, and only `DefaultFormatter` and `HumanisedFormatteter` implement this intermediate ABC.

`todoman/interactive.py` would then only take `InteractiveFormatter` as input, but prohibit `JSONFormatter`.

PR #611, opened 2026-06-06
Add categories command to prints all categories

ThePayyavula:
Introduce a new CLI command `categories` that prints all categories.


ThePayyavula:
Sorry, I made some mistakes. Will fix and get back.

PR #613, opened 2026-06-06
Solving: todo categories should show all categories  #552

ThePayyavula:
Added the command todo categories which should print all categories


ThePayyavula:
Mistakes

PR #614, opened 2026-06-06
Solved: [todo categories should show all categories  #552]

ThePayyavula:
Added the todo categories that should show all categories


todoman/cli.py
```diff
@@ -687,6 +687,17 @@ def lists(ctx: AppContext) -> None:
...
+@cli.command()
+@pass_ctx
+@catch_errors
+def categories(ctx: AppContext) -> None:
+    """Returns all the categories"""
+    categories = ctx.db.categories()
+    porcelain = ctx.formatter_class == formatters.PorcelainFormatter
+    text = json.dumps(categories) if porcelain else "\n".join(categories)
+    click.echo(text)
+
```
WhyNotHugo, line 699: `DefaultFormatter` and `PorcelainFormatter` should both implement a `format_categories` function which does the formatting, and here we just call `click.echo(ctx.formatter.format_categories(categories))`.

See how the `list_command` function is implemented below for reference.

WhyNotHugo (changes requested):
Implementation should follow the same convention as existing commands, but the idea of this new command sounds good 👍

PR #615, opened 2026-07-20
Allow configuration of multiple paths

nmeum:
For my work machine, I want to be able to access both my private and work-related to-dos. Unfortunately, they are stored in two entirely different directories, which I cannot easily match using a single `glob(3)` expression. For this kind of use case, this PR introduces a new configuration variable called `paths` which can be set to *a list of `glob(3)` expressions* to include multiple directories with `.ics` files.

The proposed patch is backwards compatible with the previous configuration format as the list's value defaults to a single-element list containing the value of the previous `path` configuration option, if the `paths` option is unset.

I intentionally tried to implement this in a way that retains backwards compatibility, I would also be open to other designs. For example, making the existing `path` configuration option a list of strings instead of a string. I haven't added tests yet but can do so once we agreed on a design.

### DIFF
--- a/todoman/cli.py
+++ b/todoman/cli.py
@@ -21,7 +21,9 @@ import click_log
 
 from todoman import exceptions
 from todoman import formatters
+from todoman.configuration import CONFIG_SPEC
 from todoman.configuration import ConfigurationError
+from todoman.configuration import expand_path
 from todoman.configuration import load_config
 from todoman.interactive import TodoEditor
 from todoman.model import Database
@@ -205,6 +207,15 @@ def validate_status(ctx: AppContext, param: click.Parameter, val: str) -> str:
     return val
 
 
+def glob_path(path: str) -> list[str]:
+    paths = [
+        path
+        for path in glob.iglob(path)
+        if isdir(path) and not path.endswith("__pycache__")
+    ]
+    return paths
+
+
 def _todo_property_options(command: Callable) -> Callable:
     click.option(
         "--category",
@@ -360,13 +371,23 @@ def cli(
     elif colour == "never":
         click_ctx.color = False
 
-    paths = [
-        path
-        for path in glob.iglob(ctx.config["path"])
-        if isdir(path) and not path.endswith("__pycache__")
-    ]
+    # XXX: There is no good way to check if an option has been explicitly set.
+    default_path = expand_path(CONFIG_SPEC[0].default)
+    if ctx.config["path"] != default_path and len(ctx.config["paths"]) != 0:
+        raise ConfigurationError("Both 'path' and 'paths' is set, use one.")
+
+    # If "paths" is unset, fallback to the old "path" config option.
+    paths = ctx.config["paths"]
     if len(paths) == 0:
-        raise exceptions.NoListsFoundError(ctx.config["path"])
+        paths = glob_path(ctx.config["path"])
+        if len(paths) == 0:
+            raise exceptions.NoListsFoundError(ctx.config["path"])
+    else:
+        paths = [p for x in paths for p in glob_path(x)]
+        if len(paths) == 0:
+            raise exceptions.NoListsFoundError(ctx.config["paths"])
+        elif len(paths) != len(set(paths)):
+            raise exceptions.DuplicatedPathError(ctx.config["paths"])
 
     ctx.db = Database(paths, ctx.config["cache_path"])
     click_ctx.call_on_close(ctx.db.close)
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1046,6 +1046,56 @@ def test_porcelain_precedence(runner: CliRunner, tmpdir: py.path.local) -> None:
     assert mocked_formatter.call_count == 1
 
 
+def test_path_and_paths_entry(
+    tmpdir: py.path.local,
+    runner: CliRunner,
+    create: Callable,
+    config: py.path.local,
+) -> None:
+    config.write(f'path = "{tmpdir}"\n', "a")
+    config.write(f'paths = ["{tmpdir}"]\n', "a")
+
+    result = runner.invoke(cli, ["list"])
+    assert result.exception
+    assert "Both 'path' and 'paths' is set, use one." in str(result.exception)
+
+
+def test_duplicated_expanded_directories(
+    tmpdir: py.path.local,
+    runner: CliRunner,
+    create: Callable,
+    config: py.path.local,
+) -> None:
+    parent = tmpdir.dirname
+    config.write(f'paths = ["{tmpdir}", "{parent}/*"]\n')
+
+    result = runner.invoke(cli, ["list"])
+    assert result.exception
+    assert result.exit_code == exceptions.DuplicatedPathError.EXIT_CODE
+
+
+def test_list_multiple_paths(
+    tmpdir: py.path.local,
+    runner: CliRunner,
+    create: Callable,
+    config: py.path.local,
+) -> None:
+    tmpdir.join("dir1").mkdir()
+    tmpdir.join("dir2").mkdir()
+    config.write(f'paths = ["{tmpdir.join("dir1")}", "{tmpdir.join("dir2")}"]\n')
+
+    result = runner.invoke(cli, ["list"], catch_exceptions=False)
+    assert not result.exception
+
+    create("test.ics", f"UID:{uuid4()}\nSUMMARY:from1\n", "dir1")
+    create("test.ics", f"UID:{uuid4()}\nSUMMARY:from2\n", "dir2")
+
+    result = runner.invoke(cli, ["list"])
+    assert not result.exception
+    assert "from1" in result.output
+    assert "from2" in result.output
+
+
 def test_duplicate_list(tmpdir: py.path.local, runner: CliRunner) -> None:
     tmpdir.join("personal1").mkdir()
     with tmpdir.join("personal1").join("displayname").open("w") as f:
--- a/todoman/configuration.py
+++ b/todoman/configuration.py
@@ -19,6 +19,10 @@ def expand_path(path: str) -> str:
     return os.path.expanduser(os.path.expandvars(path))
 
 
+def expand_paths(paths: list[str]) -> list[str]:
+    return list(map(expand_path, paths))
+
+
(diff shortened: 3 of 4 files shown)


nmeum:
Not a Python person myself, but I assume the mypy error on the CI are not caused by my changes?

<details>
<summary>mypy error output on the build.sr.ht/alpine-py312 CI.</summary>
<pre>
lint: commands[2]> mypy .
todoman/widgets.py:70: error: No return value expected  [return-value]
todoman/widgets.py:70: error: Argument 1 to "keypress" of "Edit" has incompatible type "tuple[int, int] | tuple[int] | tuple[()]"; expected "tuple[int]"  [arg-type]
todoman/widgets.py:123: error: "Widget" has no attribute "_loop"  [attr-defined]
todoman/widgets.py:150: error: Cannot infer value of type parameter "WrappedWidget" of "__init__" of "WidgetWrap"  [misc]
todoman/interactive.py:59: error: Incompatible types in assignment (expression has type "Text", variable has type "str")  [assignment]
todoman/interactive.py:60: error: List item 0 has incompatible type "tuple[int, str]"; expected "Widget | tuple[Literal['pack', WHSettings.PACK] | int, Widget] | tuple[Literal['given', WHSettings.GIVEN], int, Widget] | tuple[Literal['weight', WHSettings.WEIGHT], int | float, Widget]"  [list-item]
todoman/interactive.py:60: error: List item 1 has incompatible type "object"; expected "Widget | tuple[Literal['pack', WHSettings.PACK] | int, Widget] | tuple[Literal['given', WHSettings.GIVEN], int, Widget] | tuple[Literal['weight', WHSettings.WEIGHT], int | float, Widget]"  [list-item]
todoman/interactive.py:63: error: Argument 1 to "Pile" has incompatible type "list[tuple[str, Columns]]"; expected "Iterable[Widget | tuple[Literal['pack', WHSettings.PACK] | int, Widget] | tuple[Literal['given', WHSettings.GIVEN], int, Widget] | tuple[Literal['weight', WHSettings.WEIGHT], int | float, Widget]]"  [arg-type]
todoman/interactive.py:77: error: Argument "parent" to "ExtendedEdit" has incompatible type "TodoEditor"; expected "Widget"  [arg-type]
todoman/interactive.py:81: error: Argument "parent" to "ExtendedEdit" has incompatible type "TodoEditor"; expected "Widget"  [arg-type]
todoman/interactive.py:86: error: Argument "parent" to "ExtendedEdit" has incompatible type "TodoEditor"; expected "Widget"  [arg-type]
todoman/interactive.py:90: error: Argument "parent" to "ExtendedEdit" has incompatible type "TodoEditor"; expected "Widget"  [arg-type]
todoman/interactive.py:94: error: Argument "parent" to "ExtendedEdit" has incompatible type "TodoEditor"; expected "Widget"  [arg-type]
todoman/interactive.py:98: error: Argument "parent" to "ExtendedEdit" has incompatible type "TodoEditor"; expected "Widget"  [arg-type]
todoman/interactive.py:103: error: Argument "parent" to "PrioritySelector" has incompatible type "TodoEditor"; expected "Widget"  [arg-type]
todoman/interactive.py:143: error: "ListWalker" has no attribute "contents"  [attr-defined]
todoman/interactive.py:144: error: "ListWalker" has no attribute "contents"  [attr-defined]
todoman/interactive.py:146: error: "ListWalker" has no attribute "contents"  [attr-defined]
todoman/interactive.py:156: error: Incompatible types in assignment (expression has type "MainLoop", variable has type "None")  [assignment]
todoman/interactive.py:159: error: Argument "unhandled_input" to "MainLoop" has incompatible type "Callable[[str], None]"; expected "Callable[[str | tuple[str, int, int, int]], bool | None] | None"  [arg-type]
tests/test_widgets.py:16: error: Argument 1 to "ExtendedEdit" has incompatible type "None"; expected "Widget"  [arg-type]
tests/test_widgets.py:44: error: Argument 1 to "ExtendedEdit" has incompatible type "None"; expected "Widget"  [arg-type]
tests/test_widgets.py:60: error: Argument 1 to "ExtendedEdit" has incompatible type "None"; expected "Widget"  [arg-type]
tests/test_widgets.py:82: error: Argument 1 to "ExtendedEdit" has incompatible type "None"; expected "Widget"  [arg-type]
tests/test_widgets.py:98: error: Argument 1 to "ExtendedEdit" has incompatible type "None"; expected "Widget"  [arg-type]
tests/test_widgets.py:114: error: Argument 1 to "ExtendedEdit" has incompatible type "None"; expected "Widget"  [arg-type]
tests/test_widgets.py:168: error: Argument 1 to "PrioritySelector" has incompatible type "None"; expected "Widget"  [arg-type]
tests/test_ui.py:120: error: "ListWalker" has no attribute "contents"  [attr-defined]
tests/test_ui.py:124: error: "ListWalker" has no attribute "contents"  [attr-defined]
tests/test_ui.py:155: error: "ListWalker" has no attribute "contents"  [attr-defined]
Found 30 errors in 4 files (checked 24 source files)
</pre>
</details>

With respect to the CI coverage failure, would address this by adding tests if we agree on this API (see the original description for possible alternative designs).

todoman/cli.py
```diff
@@ -205,6 +206,15 @@ def validate_status(ctx: AppContext, param: click.Parameter, val: str) -> str:
     return val
 
 
+def glob_path(path: str) -> list[str]:
+    paths = [
+        path
+        for path in glob.iglob(path)
+        if isdir(path) and not path.endswith("__pycache__")
```
WhyNotHugo, line 214: Ignoring `__pycache__` specifically is unusual, but acceptable I guess.

todoman/cli.py
```diff
@@ -360,13 +370,17 @@ def cli(
...
     if len(paths) == 0:
-        raise exceptions.NoListsFoundError(ctx.config["path"])
+        paths = glob_path(ctx.config["path"])
+        if len(paths) == 0:
+            raise exceptions.NoListsFoundError(ctx.config["path"])
+    else:
+        init: list[str] = []
+        paths = reduce(lambda acc, x: glob_path(x) + acc, paths, init)
+        if len(paths) == 0:
+            raise exceptions.NoListsFoundError(ctx.config["paths"])
```
WhyNotHugo, line 388: If multiple paths contain directories with the same name, we should bail — this would imply we have two lists with the same name (which was previously not possible).

todoman/cli.py
```diff
@@ -360,13 +370,17 @@ def cli(
...
-    paths = [
-        path
-        for path in glob.iglob(ctx.config["path"])
-        if isdir(path) and not path.endswith("__pycache__")
-    ]
+    # If "paths" is unset, fallback to the old "path" config option.
+    paths = ctx.config["paths"]
     if len(paths) == 0:
-        raise exceptions.NoListsFoundError(ctx.config["path"])
+        paths = glob_path(ctx.config["path"])
```
WhyNotHugo, line 382: Shouldn't we update `ctx.config["paths"]` with this new value?

todoman/cli.py
```diff
@@ -360,13 +370,17 @@ def cli(
...
+        if len(paths) == 0:
+            raise exceptions.NoListsFoundError(ctx.config["path"])
+    else:
+        init: list[str] = []
+        paths = reduce(lambda acc, x: glob_path(x) + acc, paths, init)
```
WhyNotHugo: This is simpler and more efficient:

```suggestion
        paths = [p for x in paths for p in glob_path(x)]
```

WhyNotHugo (commented):
Overall idea sounds good. Note a *huge* fan, but I don't see a cleaner way to support this, and the request sounds reasonable.

I think we should fail with an error when both `path` and `paths` are set.

Mostly minor comments, general changes are good.

todoman/cli.py
nmeum: That was part of the existing implementation, hence I retained it.

https://github.com/pimutils/todoman/blob/78e35788606a0add0c4f99ceb702c3ea0d0b1346/todoman/cli.py#L366

todoman/cli.py
nmeum: Added a check for this.

https://github.com/pimutils/todoman/blob/91c21b3a2f42ba78353aaf0eaf83218a6e05bcc3/todoman/cli.py#L389-L390

todoman/cli.py
nmeum: Thanks, applied!

todoman/cli.py
nmeum: Is there a specific reason why this should be necessary? The existing implementation does also not update `ctx.config["path"]` after calling `.iglob()` on it:

https://github.com/pimutils/todoman/blob/78e35788606a0add0c4f99ceb702c3ea0d0b1346/todoman/cli.py#L363-L369

nmeum:
> Overall idea sounds good. Note a huge fan, but I don't see a cleaner way to support this, and the request sounds reasonable.

Thanks for the feedback! This could get a little cleaner by deprecating the `path` configuration option in the future. I think having a list of glob expressions instead of a single glob expressions for the path configuration is generally a "cleaner" approach to configuring searched directories.

todoman/cli.py
```diff
@@ -360,13 +371,23 @@ def cli(
...
-    paths = [
-        path
-        for path in glob.iglob(ctx.config["path"])
-        if isdir(path) and not path.endswith("__pycache__")
-    ]
+    # XXX: There is no good way to check if an option has been explicitly set.
+    default_path = expand_path(CONFIG_SPEC[0].default)
+    if ctx.config["path"] != default_path and len(ctx.config["paths"]) != 0:
+        raise ConfigurationError("Both 'path' and 'paths' is set, use one.")
+
```
nmeum, line 378: > I think we should fail with an error when both `path` and `paths` are set.

Implemented this here, but it's a little hacky as `path` has a default value and we thus need to determine if it has been explicitly set by the user, for which there doesn't seem to be an API at the moment.

nmeum: Also can't do this via a configuration validation function as it doesn't have access to the already parsed options.

nmeum:
As an alternative, potentially less hacky design, we could also make `path` an `Any` type and then use the old behavior if it is a string, or new one if it is a list.

